PR #42, opened 2019-10-09
[X-Address] Consider X-Addresses as valid address

keefertaylor:
Update the implementation of `Utils` class to work with x-addresses. The new implementation doesn't distinguish between x-address and classic addresses as I don't see an obvious use case for why a Xpring SDK user want's this functionality.

Changes:
- Updates the underlying `ripple-address-codec` dependency to 4.0.0, which has x-address support. 
- Updates the implementation of `Utils.isValidAddress` for breaking changes in dependency update and to validate x-addresses
- [Cleanup] Add missing documentation in other utils

New APIs:
N/A

Breaking Changes:
N/A

See Also: 
X-Address Format: https://xrpaddress.info/

### DIFF
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -5,20 +5,23 @@ var addressCodec = require("ripple-address-codec");
 class Utils {
   /**
    * Validate that the given string is a valid address on the XRP Ledger.
-   * 
+   *
    * This function returns true for both x-addresses and classic addresses.
    * @see https://xrpaddress.info/
-   * 
+   *
    * @param address An address to check.
    * @returns True if the address is valid, otherwise false.
    */
   public static isValidAddress(address: string): boolean {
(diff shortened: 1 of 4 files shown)


src/utils.ts
```diff
@@ -3,14 +3,38 @@
 var addressCodec = require("ripple-address-codec");
 
 class Utils {
+  /**
+   * Validate that the given string is a valid address on the XRP Ledger.
+   *
+   * This function returns true for both x-addresses and classic addresses.
```
intelliot: nit: I think I prefer to use capitalize the X

test/utils-test.ts
```diff
@@ -14,6 +14,17 @@ describe("utils", function(): void {
     assert.isTrue(validAddress);
   });
 
+  it("isValidAddress() - Valid X Address", function(): void {
+    // GIVEN a valid x-address.
```
intelliot: same nit (capitalize X)

PR #43, opened 2019-10-09
[X-Address] Add Encoding and Decoding of X-Addresses

keefertaylor:
Add methods on `Utils` class to encode and decode x-addresses. 

Note minor shims are required to work with the underlying library. The underlying library prefers `string|false` while Xpring Common JS prefers `string|undefined`. `undefined` is preferred because it works well with strongly typed languages (ie. `String` in java is `String` of `null`) and optional monads (i.e. `String` in swift is always `String` and `String?` is `String` or `nil`).

Changes:
- Add new APIs to `Utils` class
- Add new `ClassicAddress` interface representing the property bag of objects returned from x-address decoding.

New APIs:
- `Utils.encodeXAddress(address: string, tag: number|undefined): string|undefined`
- `Utils.decodeXAddres(xAddress: string): ClassicAddress|undefined`

Breaking Changes:
N/A

See Also: 
X-Address Format: https://xrpaddress.info/

### DIFF
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -20,6 +20,25 @@ class Utils {
   }
 
   /**
+   * Encode the given classic address and tag into an x-address.
+   * 
+   * @see https://xrpaddress.info/
+   * 
+   * @param classicAddress A classic address to encode.
+   * @param tag An optional tag to encode.
+   * @returns A new x-address if inputs were valid, otherwise undefined.
+   */
+  public static encodeXAddress(classicAddress: string, tag: number|undefined): string|undefined {
+    if (!addressCodec.isValidClassicAddress(classicAddress)) {
+      return undefined;
+    }
+
+    // Xpring Common JS's API takes a string|undefined while the underlying address library wants string|false.
+    const shimTagParameter = tag != undefined ? tag : false;
+    return addressCodec.classicAddressToXAddress(classicAddress, shimTagParameter)
+  }
+
+  /**
    * Convert the given byte array to a hexadecimal string.
    *
    * @param bytes An array of bytes
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -10,7 +10,7 @@ export interface ClassicAddress {
   address: string;
 
   /** An optional tag. */
-  tag: number | undefined;
+  tag?: number;
 }
 
 class Utils {
@@ -48,7 +48,7 @@ class Utils {
     }
 
     // Xpring Common JS's API takes a string|undefined while the underlying address library wants string|false.
-    const shimTagParameter = tag != undefined ? tag : false;
+    const shimTagParameter = tag !== undefined ? tag : false;
     return addressCodec.classicAddressToXAddress(
       classicAddress,
       shimTagParameter
@@ -71,7 +71,7 @@ class Utils {
     let shimClassicAddress = addressCodec.xAddressToClassicAddress(xAddress);
     return {
       address: shimClassicAddress.classicAddress,
-      tag: shimClassicAddress.tag != false ? shimClassicAddress.tag : undefined
+      tag: shimClassicAddress.tag !== false ? shimClassicAddress.tag : undefined
     };
   }
 
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -2,6 +2,17 @@
 
 var addressCodec = require("ripple-address-codec");
 
+/**
+ * A simple property bag which contains components of a classic address. Components contained in this object are neither sanitized or validated.
+ */
+export interface ClassicAddress {
+  /** A classic address. */
+  address: string;
+
+  /** An optional tag. */
+  tag: number | undefined;
+}
+
 class Utils {
   /**
    * Validate that the given string is a valid address on the XRP Ledger.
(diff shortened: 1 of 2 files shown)


src/utils.ts
```diff
@@ -19,6 +30,51 @@ class Utils {
...
+   */
+  public static decodeXAddress(xAddress: string): ClassicAddress | undefined {
+    if (!addressCodec.isValidXAddress(xAddress)) {
+      return undefined;
+    }
+
+    let shimClassicAddress = addressCodec.xAddressToClassicAddress(xAddress);
+    return {
+      address: shimClassicAddress.classicAddress,
+      tag: shimClassicAddress.tag != false ? shimClassicAddress.tag : undefined
```
Stormtv: Is there a reason here to have "!=" instead of "!=="? If `tag === "0"` or `tag === 0` that is equivalent to false when using `!=`

src/utils.ts
```diff
@@ -19,6 +30,51 @@ class Utils {
...
+   *
+   * @see https://xrpaddress.info/
+   *
+   * @param classicAddress A classic address to encode.
+   * @param tag An optional tag to encode.
+   * @returns A new x-address if inputs were valid, otherwise undefined.
+   */
+  public static encodeXAddress(
+    classicAddress: string,
+    tag: number | undefined
```
Stormtv: instead of defining it as `tag: number | undefined` should we instead use the optional parameter syntax of `tag?: number`?

src/utils.ts
```diff
@@ -2,6 +2,17 @@
...
 
+/**
+ * A simple property bag which contains components of a classic address. Components contained in this object are neither sanitized or validated.
+ */
+export interface ClassicAddress {
+  /** A classic address. */
+  address: string;
+
+  /** An optional tag. */
+  tag: number | undefined;
```
Stormtv: instead of defining it as tag: number | undefined should we instead use the optional parameter syntax of tag?: number?

src/utils.ts
```diff
@@ -19,6 +30,51 @@ class Utils {
...
+  public static encodeXAddress(
+    classicAddress: string,
+    tag: number | undefined
+  ): string | undefined {
+    if (!addressCodec.isValidClassicAddress(classicAddress)) {
+      return undefined;
+    }
+
+    // Xpring Common JS's API takes a string|undefined while the underlying address library wants string|false.
+    const shimTagParameter = tag != undefined ? tag : false;
```
Stormtv: use `!==` instead of `!=`

intelliot:
On the question of `false` vs. `undefined`, I actually went back and forth on that several times. I think one of the reasons I settled on `false` is that, semantically, the X-address is not just saying that the tag is undefined (not known or not set yet), but that _no tag should be used_. On the other hand, it should be possible to intentionally "add/change" the tag of a given X-address; I'm not sure if there's demand for that feature, but it would be straightforward to add.

This is not a big deal to me though, and I think it's completely fine to use `undefined` here; especially if that makes more sense in other languages.

keefertaylor:
Re Elliot: I think that in order to map this into higher order languages you can't have polymorphic values for a parameter (ie. you can't have string|bool in most languages).  The only way to accomplish this would be to use a `Object` reference in Java or an `Any` reference in Swift, which is definitely an antipattern in those languages. 

In general, I'd suggest that polymorphic variables are a lot harder to reason about for programmers (This variable is a "string or undefined" is different than "This variable is a defined string or a singular bool value). Agree that this case is relatively simple and not worth bikeshedding over.

src/utils.ts
keefertaylor: This is cool, I did not realize you could mark params as optional this way. 

I think this is a poor choice for this particular API because it means that TS programmers can unintentionally omit this param. I think that all inputs to an X-address should be explicitly stated, since X-addresses are specifically trying to solve the problem of people leaving off tags. 

If you feel this is a strong anti-pattern in TS / JS LMK and I don't care too much .

src/utils.ts
keefertaylor: Done

src/utils.ts
keefertaylor: Done

keefertaylor:
Also as a sanity check, I don't need to worry about overflow here, right? 

JavaScript is using 64 bit floating point which should give me accuracy [up to 999,999,999,999,999,999](https://www.w3schools.com/js/js_numbers.asp), while XRP tags are 32 bit signed integers (2^32 = 4,294,967,296 < 999,999,999,999,999,999).

src/utils.ts
Stormtv: Interesting, I think it is okay to do `number | undefined` but `tag?: number` is the same thing technically.

Why is undefined an option if we are forcing people to always have a tag?

Is there a default value for destination tag?

Stormtv: NVM talked on slack I see the reasoning now

intelliot:
> JavaScript is using 64 bit floating point which should give me accuracy up to 999,999,999,999,999,999, while XRP tags are 32 bit signed integers (2^32 = 4,294,967,296 < 999,999,999,999,999,999).

The JS `number` type's largest exact integral value is 2^53 - 1, or only 9,007,199,254,740,991.

XRP tags are 32-bit **unsigned** integers. The largest value is 2^32 - 1.

Yes, this can always be safely stored in a `number`.


We've had a lot of discussion around supporting 64-bit tags, but JS's lack of a 64-bit unsigned integer type is one of the reasons we haven't prioritized it. [BigInt](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/BigInt) is coming soon, but it's not yet supported by major browsers like Edge/Safari. (It was added in Node.js 10.4.) Regardless, if we did support 64-bit tags, I think that using strings would be an easier/safer way to do it anyway (in addition to having wider compatibility).

Another reason 64-bit tags aren't necessary is that we recommend that users rotate tags (and decommission old ones) on a regular basis, for privacy reasons. If this is practiced, then 32 bits should be enough.

PR #45, opened 2019-10-15
Implement X-Address functionality in Serializer.

keefertaylor:
Fields ii a Payment protocol buffer could either be a classic address or an X-address.  This PR decodes and correctly handles the X-Address case. 

Note that for simplicity, I've chosen to not implement a `DestinationTag` in the underlying protocol buffers. Instead, Payments which want to use a `DestinationTag` will have to encode their own X-addresses. In practice, the end user's library (Xpring-JS, XpringKit, etc) will do this for them and they will be none the wiser. 

Added unit tests as well.

### DIFF
--- a/src/serializer.ts
+++ b/src/serializer.ts
@@ -83,7 +83,7 @@ class Serializer {
       delete json.DestinationTag;
     } else {
       json.Destination = decodedXAddress.address;
-      if (decodedXAddress.tag) {
+      if (decodedXAddress.tag !== undefined) {
         json.DestinationTag = decodedXAddress.tag;
       }
     }
--- a/src/serializer.ts
+++ b/src/serializer.ts
@@ -3,9 +3,17 @@ import { Transaction } from "../generated/transaction_pb";
 import { FiatAmount } from "../generated/fiat_amount_pb";
 import { XRPAmount } from "../generated/xrp_amount_pb";
 import { Currency } from "../generated/currency_pb";
+import Utils from "./utils";
 
 /* Allow `any` since this class doing progressive conversion of protocol buffers to JSON. */
 /* eslint-disable @typescript-eslint/no-explicit-any */
+
+interface PaymentJSON {
+  Amount: object | string;
+  Destination: string;
+  DestinationTag?: number | undefined;
+  TransactionType: string;
+}
 
 /**
  * Provides functionality to serialize from protocol buffers to JSON objects.
@@ -62,11 +70,22 @@ class Serializer {
    * @returns {Object} The Payment as JSON.
    */
   private static paymentToJSON(payment: Payment): object | undefined {
-    const json = {
-      TransactionType: "Payment",
-      Destination: payment.getDestination(),
-      Amount: {}
+    const json: PaymentJSON = {
+      Amount: {},
+      Destination: "",
+      DestinationTag: 0,
+      TransactionType: "Payment"
     };
+
+    // If an x-address was able to be decoded, add the components to the json.
+    const decodedXAddress = Utils.decodeXAddress(payment.getDestination());
+    if (decodedXAddress === undefined) {
+      json.Destination = payment.getDestination();
+      delete json.DestinationTag;
+    } else {
+      json.Destination = decodedXAddress.address;
+      json.DestinationTag = decodedXAddress.tag;
+    }
 
     const amountCase = payment.getAmountCase();
     switch (amountCase) {
--- a/src/serializer.ts
+++ b/src/serializer.ts
@@ -11,7 +11,7 @@ import Utils from "./utils";
 interface PaymentJSON {
   Amount: object | string;
   Destination: string;
-  DestinationTag?: number | undefined;
+  DestinationTag?: number;
   TransactionType: string;
 }
 
--- a/src/serializer.ts
+++ b/src/serializer.ts
@@ -79,7 +79,7 @@ class Serializer {
 
     // If an x-address was able to be decoded, add the components to the json.
     const decodedXAddress = Utils.decodeXAddress(payment.getDestination());
-    if (decodedXAddress === undefined) {
+    if (!decodedXAddress) {
       json.Destination = payment.getDestination();
       delete json.DestinationTag;
     } else {
--- a/src/serializer.ts
+++ b/src/serializer.ts
@@ -73,7 +73,6 @@ class Serializer {
     const json: PaymentJSON = {
       Amount: {},
       Destination: "",
-      DestinationTag: 0,
       TransactionType: "Payment"
     };
 
@@ -84,7 +83,9 @@ class Serializer {
       delete json.DestinationTag;
     } else {
       json.Destination = decodedXAddress.address;
-      json.DestinationTag = decodedXAddress.tag;
+      if (decodedXAddress.tag) {
+        json.DestinationTag = decodedXAddress.tag;
+      }
     }
 
     const amountCase = payment.getAmountCase();
--- a/test/serializer-test.ts
+++ b/test/serializer-test.ts
@@ -40,7 +40,7 @@ describe("serializer", function(): void {
     // THEN the result is as expected.
     const expectedJSON = {
       Account: account,
-      Amount: value + "",
+      Amount: value.toString(),
       Destination: destination,
       Fee: fee,
       Sequence: sequence,
@@ -82,7 +82,7 @@ describe("serializer", function(): void {
     // THEN the result is as expected.
     const expectedJSON = {
       Account: account,
-      Amount: value + "",
+      Amount: value.toString(),
       Destination: "rU6K7V3Po4snVhBBaU29sesqs2qTQJWDw1",
       DestinationTag: 12345,
       Fee: fee,
@@ -125,7 +125,7 @@ describe("serializer", function(): void {
     // THEN the result is as expected.
     const expectedJSON = {
       Account: account,
-      Amount: value + "",
+      Amount: value.toString(),
       Destination: "rU6K7V3Po4snVhBBaU29sesqs2qTQJWDw1",
       DestinationTag: undefined,
       Fee: fee,
--- a/test/serializer-test.ts
+++ b/test/serializer-test.ts
@@ -127,7 +127,6 @@ describe("serializer", function(): void {
       Account: account,
       Amount: value.toString(),
       Destination: "rU6K7V3Po4snVhBBaU29sesqs2qTQJWDw1",
-      DestinationTag: undefined,
       Fee: fee,
       Sequence: sequence,
       TransactionType: "Payment",
--- a/test/serializer-test.ts
+++ b/test/serializer-test.ts
@@ -9,7 +9,7 @@ import "mocha";
 
 describe("serializer", function(): void {
   it("serializes a payment in XRP", function(): void {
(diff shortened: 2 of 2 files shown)


src/serializer.ts
```diff
@@ -62,12 +70,23 @@ class Serializer {
...
+    const json: PaymentJSON = {
+      Amount: {},
+      Destination: "",
+      DestinationTag: 0,
+      TransactionType: "Payment"
     };
 
+    // If an x-address was able to be decoded, add the components to the json.
+    const decodedXAddress = Utils.decodeXAddress(payment.getDestination());
+    if (decodedXAddress === undefined) {
```
Stormtv: NIT: I'm curious why the explicit check on undefined over just `if (!decodedXAddress) {`? Is this just a best practice type of thing? Also curious why the function decodeXAddress returns undefined vs just doing a naked return.

test/serializer-test.ts
```diff
@@ -50,6 +50,92 @@ describe("serializer", function(): void {
...
+
+    // WHEN the transaction is serialized to JSON.
+    const serialized = Serializer.transactionToJSON(transaction);
+
+    // THEN the result is as expected.
+    const expectedJSON = {
+      Account: account,
+      Amount: value + "",
+      Destination: "rU6K7V3Po4snVhBBaU29sesqs2qTQJWDw1",
+      DestinationTag: undefined,
```
Stormtv: NIT: Not sure if you need to mention the destination tag for the object here.

test/serializer-test.ts
```diff
@@ -50,6 +50,92 @@ describe("serializer", function(): void {
...
+    transaction.setPayment(payment);
+    transaction.setSigningPublicKeyHex(publicKey);
+
+    // WHEN the transaction is serialized to JSON.
+    const serialized = Serializer.transactionToJSON(transaction);
+
+    // THEN the result is as expected.
+    const expectedJSON = {
+      Account: account,
+      Amount: value + "",
```
Stormtv: NIT: casting value to a string with the + "" is the fastest way to convert an int to a string but I'm not a fan of it from syntax wise

src/serializer.ts
keefertaylor: I honestly just figured there was some weird behavior in javascript if I used this style :) done.

test/serializer-test.ts
keefertaylor: Assume `value.toString()` works, LMK if you prefer something else. 

Fixed here and elsewhere.

src/serializer.ts
keefertaylor: > Also curious why the function decodeXAddress returns undefined vs just doing a naked return.

You're suggesting that x-address would just return the classic address, if a classic address is passed as input. There are 3 cases:


| input  | As Implemented | Naked Return |
|---|---|---|
| valid classic address | undefined | input classic address |
| valid x address | undefined | decoded classic address |
|some invalid string ("xrp") | undefined | undefined (?)|

I think returning undefined is the most expected thing. Naked returns complicate the return values of this function (now 3 things can be returned). 

You could argue that I should really implement an `isXAddress` function, which might help this, but I'd still need to do something like:
```javascript
if (isXAddr(addr) && decodeXAddr(addr) != undefined) {
    // do stuff 
}
```

or 

```javascript 
if (isXAddr(xaddr)) {
   const decoded = decodeXAddr(xAddr);
   // if decoded is undefined, try some error handling
}
```

test/serializer-test.ts
keefertaylor: The tests do pass. I don't know how ripple-lib will ultimately handle this.  Perhaps @intelliot  knows? 

Aside, this is one of the reasons I want to move to an mono repo. I can't test this until I upgrade Appian and Xpring-JS to use this code, without setting up `npm link`.

keefertaylor:
PTAL!

test/serializer-test.ts
intelliot: I recommend using `npm link` (or `yarn link`) -- it should work fairly well (feel free to let me know if you have issues).

Also, if you'd like to switch to a monorepo, I'm in favor. Go for it.

How is this JSON used? Technically, `undefined` is not a valid JSON value (but it is valid in JavaScript). If the value should be undefined (when parsed), then it should be omitted from the JSON.

Where/how does this get passed to ripple-lib?

src/serializer.ts
```diff
@@ -3,10 +3,18 @@ import { Transaction } from "../generated/transaction_pb";
...
 import { Currency } from "../generated/currency_pb";
+import Utils from "./utils";
 
 /* Allow `any` since this class doing progressive conversion of protocol buffers to JSON. */
 /* eslint-disable @typescript-eslint/no-explicit-any */
 
+interface PaymentJSON {
+  Amount: object | string;
+  Destination: string;
+  DestinationTag?: number | undefined;
```
intelliot: This allows `DestinationTag` to be omitted, or if it is provided, it can be a `number` or `undefined`. I think it would make more sense for it to be one of these instead:
```
DestinationTag?: number;
```
or
```
DestinationTag: number | undefined;
```

src/serializer.ts
```diff
@@ -3,10 +3,18 @@ import { Transaction } from "../generated/transaction_pb";
...
+  TransactionType: string;
```
intelliot: What `TransactionType`s are supported? If "Payment" is the only one, then I would do:
```
  TransactionType: "Payment";
```

src/serializer.ts
```diff
@@ -3,10 +3,18 @@ import { Transaction } from "../generated/transaction_pb";
 import { FiatAmount } from "../generated/fiat_amount_pb";
 import { XRPAmount } from "../generated/xrp_amount_pb";
 import { Currency } from "../generated/currency_pb";
+import Utils from "./utils";
 
 /* Allow `any` since this class doing progressive conversion of protocol buffers to JSON. */
 /* eslint-disable @typescript-eslint/no-explicit-any */
 
+interface PaymentJSON {
+  Amount: object | string;
```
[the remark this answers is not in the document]
keefertaylor: Support for Fiat payments is kinda around: 
https://github.com/xpring-eng/xpring-common-protocol-buffers/blob/master/proto/payment.proto

I think there are some optimizations that I might try to make to the internal protocol buffer structure here. 

Because of the current protocol buffer format we need to support xrp and fiat here. I'm not particularly interested in doing fiat stuff at this time - it's low priority given the other things I'm working on. 

I think there's some long term thinking about whether or not these two amounts really need to be different in the client side API  (i.e. using `string` for XRP is just shorthand for a fiat amount).  That's probably worth a longer discussion, possibly with a document and with consensus from the rippled team. 

Shorter term (this week) my goal is to just drop protocol buffers from the public APIs completely so that we can make internal refactors and changes as needed.

src/serializer.ts
keefertaylor: It's extensible, based on the `oneof` cases in the protocol buffers. At the moment, payment is the only supported option. 

I'm happy to hardcode this, but it makes the if statements more messy (because we still have to switch on a oneof) and the code is already written as is, so I'd prefer to leave it unless you have strong objections.

src/serializer.ts
keefertaylor: Done

src/serializer.ts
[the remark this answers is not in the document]
intelliot: I don't think "Fiat" is the right term here. We call them "[issued currencies](https://xrpl.org/issued-currencies-overview.html)", "issuances", or "IOUs". These issued currencies can represent anything of value, including digital currencies like Bitcoin or Ether, other assets like precious metals, or other forms of value.

For more details see:
- [Consensus Principles and Rules](https://xrpl.org/consensus-principles-and-rules.html)
- [Data Types - Currencies](https://xrpl.org/basic-data-types.html#currencies)
- [Become an XRP Ledger Gateway](https://xrpl.org/become-an-xrp-ledger-gateway.html)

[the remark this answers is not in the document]
intelliot: (By the way, one of the "improvements" that ripple-lib made on the rippled API was to always [use an object](https://xrpl.org/rippleapi-reference.html#currency) for an amount; unlike the rippled API which uses a string for amounts of XRP drops. I don't care either way -- in fact I kind of prefer rippled's way -- but I thought I'd mention this just in case.)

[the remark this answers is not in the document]
keefertaylor: Noted - I filed an issue for this in Asana. I've been thinking about restructuring this for a bit. 

The initial bit is going to be refactoring these objects out of the public API, which I'm currently working on.

src/serializer.ts
intelliot: It's fine to leave it. I don't understand this, though:

> we still have to switch on a oneof

src/serializer.ts
keefertaylor: Resolving since Tyler approved changes.

test/serializer-test.ts
keefertaylor: Resolving since Tyler approved changes.

src/serializer.ts
keefertaylor: Sorry for not explaining. 

A oneof is a set of fields where only one will have a value: https://developers.google.com/protocol-buffers/docs/proto#oneof. The generated code guarantees this invariant. 

This is used in the following way:
- The `Transaction` proto has all the [Common Fields](https://xrpl.org/transaction-common-fields.html)
- The [`Transaction` proto](https://github.com/xpring-eng/xpring-common-protocol-buffers/blob/master/proto/transaction.proto) also has a `oneof` (`transaction_data`) field which has sub-messages for every type of Transaction supported. 
- Right now, we only support [`Payment`](https://github.com/xpring-eng/xpring-common-protocol-buffers/blob/master/proto/payment.proto), so the oneof is always set to Payment
- We should switch on which field is set: https://github.com/xpring-eng/xpring-common-js/blob/master/src/serializer.ts#L43

Happy to elaborate on any part of this.

test/serializer-test.ts
keefertaylor: I apologize, I didn't actually submit my response to your query. 

I use this JSON because it's the format that other ripple libraries seem to want. For instance, to generate a signature, I need to use `ripple-binary-codec`: https://github.com/xpring-eng/xpring-common-js/blob/master/src/signer.ts#L33

When this gets submitted back to Rippled, I also encode back to JSON, encode the signature, and then use the `submit` method: 
https://github.com/xpring-eng/AppianJS/blob/master/src/index.js#L101

If ripplelib and ripplebinary codec will take x-addresses in the destination field, then I can actually drop this PR.

keefertaylor:
Sorry about that. Fixed the tag issue.

src/serializer.ts
```diff
@@ -62,12 +70,24 @@ class Serializer {
...
     };
 
+    // If an x-address was able to be decoded, add the components to the json.
+    const decodedXAddress = Utils.decodeXAddress(payment.getDestination());
+    if (!decodedXAddress) {
+      json.Destination = payment.getDestination();
+      delete json.DestinationTag;
+    } else {
+      json.Destination = decodedXAddress.address;
+      if (decodedXAddress.tag) {
```
intelliot: Since `0` is a valid `tag`, I think this should be:
```
      if (decodedXAddress.tag !== undefined) {
```

keefertaylor: Done

PR #47, opened 2019-10-18
Upgrade protocol buffers

keefertaylor:
Creates minor changes that were added when the linter was initially wired in https://github.com/xpring-eng/xpring-common-protocol-buffers/pull/11.

The currency protocol buffer was never exposed in a public or network API so it should be safe to re-number the fields (Though we should avoid this in the future).


src/serializer.ts
```diff
@@ -119,9 +119,11 @@ class Serializer {
    * @param {proto.FiatAmount.Currency} currency The Currency to convert.
    * @returns {String} The Currency as JSON.
    */
-  private static currencyToJSON(currency: Currency): string {
+  private static currencyToJSON(currency: Currency): string | undefined {
     switch (currency) {
-      case Currency.USD:
+      case Currency.CURRENCY_INVALID:
```
intelliot: Not sure what `CURRENCY_INVALID` will be used for, but if it should not occur (i.e. it represents an error) then we could consider throwing an error here.

PR #50, opened 2019-10-25
Add Validation Functions for Classic and X Addresses

keefertaylor:
These functions are needed to perform sends at the client SDK level.

### DIFF
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -39,7 +39,7 @@ class Utils {
    * @returns True if the address is a valid X-address, otherwise false.
    */
   public static isValidXAddress(address: string): string {
-    return addressCodec.isValidXAddress(address)
+    return addressCodec.isValidXAddress(address);
   }
 
   /**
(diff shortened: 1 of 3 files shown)


src/utils.ts
```diff
@@ -30,6 +30,30 @@ class Utils {
     );
   }
 
+  /**
+   * Validate if the given string is a valid X-address on the XRP Ledger.
```
intelliot: nit: prefer `Validate whether the given string is a valid X-address for the XRP Ledger.`

`on the XRP Ledger` implies that the function is querying the ledger itself to see whether the account exists there, but this is just an offline check

src/utils.ts
```diff
@@ -30,6 +30,30 @@ class Utils {
...
+   *
+   * @param address An address to check.
+   * @returns True if the address is a valid X-address, otherwise false.
+   */
+  public static isValidXAddress(address: string): string {
+    return addressCodec.isValidXAddress(address);
+  }
+
+  /**
+   * Validate if the given string is a valid classic address on the XRP Ledger.
```
intelliot: `Validate whether the given string is a valid classic address for the XRP Ledger.`

keefertaylor:
PTAL

PR #51, opened 2019-10-25
Add Utility Function for Transaction Hashing 

keefertaylor:
The gRPC API doesn't pass back the transaction hash when submitting a transaction. Instead, it returns a hex encoded blob. We need a utility to convert that to a hex, which is what our users will really want to interact with. 

Add a utility to do this. Add unit tests. 

Note: I realize that `Utils` is fast becoming a dumping ground for all sorts of things. I think that this is still small enough that it doesn't justify splitting out a `Hashing` class or the like, but happy to hear reviewer's throughts.

### DIFF
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -1,6 +1,9 @@
 "use strict";
 
-var addressCodec = require("ripple-address-codec");
+import { createHash } from 'crypto'
+
+const addressCodec = require("ripple-address-codec");
+const isHex = require("is-hex");
 
 /**
  * A simple property bag which contains components of a classic address. Components contained in this object are neither sanitized or validated.
@@ -82,7 +85,7 @@ class Utils {
    * @returns An encoded hexadecimal string.
    */
   public static toHex(bytes: Uint8Array): string {
-    return Buffer.from(bytes).toString("hex");
+    return Buffer.from(bytes).toString("hex").toUpperCase();
   }
 
   /**
@@ -93,6 +96,36 @@ class Utils {
    */
   public static toBytes(hex: string): Uint8Array {
     return Uint8Array.from(Buffer.from(hex, "hex"));
+  }
+
+  /**
+   * Convert the given transaction blob to a transaction hash.
+   * 
+   * @param transactionBlobHex A hexadecimal encoded transaction blob.
+   * @returns A hex encoded hash if the input was valid, otherwise undefined.
+   */
+  public static transactionBlobToTransactionHash(transactionBlobHex: string): string|undefined {
+    if (!isHex(transactionBlobHex)) {
+      return undefined
+    }
+
+    const prefixedTransactionBlob = this.toBytes("54584E00" + transactionBlobHex);
+    const hash = this.sha512Half(prefixedTransactionBlob);
+    return this.toHex(hash);
+  }
+
+  /**
+   * Compute the SHA512 half hash of the given bytes.
+   *
+   * @param input The input to hash.
+   * @returns The hash of the input.
+   */
+  private static sha512Half(bytes: Uint8Array): Uint8Array {
+    const sha512 = createHash("sha512");
+    const hashHex = sha512.update(bytes).digest('hex').toUpperCase();
+    const hash = this.toBytes(hashHex);
+
+    return hash.slice(0, hash.length / 2);
   }
 }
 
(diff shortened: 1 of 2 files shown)


src/utils.ts
```diff
@@ -1,6 +1,9 @@
 "use strict";
 
-var addressCodec = require("ripple-address-codec");
+import { createHash } from "crypto";
```
intelliot: I'm not sure if "crypto" is available on all the platforms that this code will run on? (I think it's a Node.js module.) In ripple-lib, I believe we use [create-hash](https://github.com/crypto-browserify/createHash). Just a thought.

src/utils.ts
```diff
@@ -94,6 +99,45 @@ class Utils {
...
+   */
+  public static transactionBlobToTransactionHash(
+    transactionBlobHex: string
+  ): string | undefined {
+    if (!isHex(transactionBlobHex)) {
+      return undefined;
+    }
+
+    const prefixedTransactionBlob = this.toBytes(
+      "54584E00" + transactionBlobHex
```
intelliot: Maybe put the prefix into a constant

src/utils.ts
Stormtv: It is a node.js module but it should be polyfilled by webpack for the distribution.

src/utils.ts
```diff
@@ -94,6 +99,45 @@ class Utils {
...
+  private static sha512Half(bytes: Uint8Array): Uint8Array {
+    const sha512 = createHash("sha512");
+    const hashHex = sha512
+      .update(bytes)
+      .digest("hex")
+      .toUpperCase();
+    const hash = this.toBytes(hashHex);
+
+    /* eslint-disable @typescript-eslint/no-magic-numbers */
+    return hash.slice(0, hash.length / 2);
```
Stormtv: NIT
I believe a sha512 hash always has a length of 64 so you could just write 32 instead of length / 2

keefertaylor: done

src/utils.ts
keefertaylor: I'd rather rely on a core module or a webpack polyfill than a public library. 

It looks like ripple-lib uses the core crypto module here: https://github.com/ripple/ripple-lib/blob/develop/src/common/hashes/sha512Half.ts#L1

keefertaylor:
> > The rippled API only returns us a transaction blob when a transaction is submitted.
> 
> I believe the rippled API actually gives us the transaction hash (aka id) in the response, in `result.tx_json.hash`. It gives us other info about the tx as well. [Docs here](https://xrpl.org/submit.html)

You are correct. I'm mostly only interested in the hash at this point, although we do send back a few other fields across gRPC at the moment:
https://github.com/xpring-eng/xpring-common-protocol-buffers/blob/master/proto/submit_signed_transaction_response.proto

As gRPC becomes productionized in rippled, we'd expect to get the full set of fields.

src/utils.ts
keefertaylor: done

intelliot:
This is good to have, but I think the first line of the PR description should be amended, since the hash (in hex) is provided in the rippled API response

keefertaylor:
I just re-looked at the responses. It is provided, under the transaction object. 

Outside of the scope of this PR, It's really confusing how 'Transaction' can mean different things, but maybe I misunderstand:
- When formulating a raw JSON transaction, we call this a 'Transaction'
- When signing, we insert the signature into the previous 'Transaction' JSON, we also call this a 'Transaction'
- When submitting, rippled gives us back a 'Transaction' which now has the signature *and* hash attached. 

Am I right that we call all 3 of these objects 'transactions'? The scheme wherein different fields are populated by what stage of the lifecycle the transaction is in seems confusing? Is there a better mental model I should be using here?

intelliot:
It's the same Transaction at different stages of its lifecycle, and it also has a different form/fields in each stage.

If you think it's beneficial, though, feel free to change the names. These names are already in use in some places:
- Transaction JSON
- txJSON
- signedTransaction
- Transaction response

keefertaylor:
I think that changing names probably just makes things more confusing and until there's an API versioning strategy we'd just be needlessly breaking folks.  I mostly wanted to make sure my mental model is correct.  If / when we revisit this API I have thoughts, but that will be in the future. 

The issue here is really that my gRPC API isn't populating all the fields returned from rippled then. I think that this will resolve at some point in the future, at which point this code can go away. In the short term, I do think that it is useful though. 

I amended the PR description, as requested.

PR #53, opened 2019-11-05
Upgrade Signer / Serializer to handle X-Addresses

keefertaylor:


src/serializer.ts
```diff
@@ -32,10 +32,31 @@ class Serializer {
...
     // Convert fields names where direct conversion is possible.
-    this.convertPropertyName("account", "Account", object);
     this.convertPropertyName("sequence", "Sequence", object);
     this.convertPropertyName("signingPublicKeyHex", "SigningPubKey", object);
 
+    // Convert account field, handling X-Addresses if needed.
+    const account = transaction.getAccount();
+    if (account == undefined || !Utils.isValidAddress(account)) {
+      return undefined;
+    }
```
Stormtv: Nitpicking here but I usually would write this as follows
```
if (!account || !Utils.isValidAddress(account)) return
```

src/serializer.ts
```diff
@@ -32,10 +32,31 @@ class Serializer {
...
+    var normalizedAccount = account;
+    if (Utils.isValidXAddress(account)) {
+      const decodedClassicAddress = Utils.decodeXAddress(account);
+      if (decodedClassicAddress == undefined) {
+        return undefined;
+      }
```
Stormtv: Same as above
```
if (!decodedClassicAddress) return
```

src/serializer.ts
```diff
@@ -32,10 +32,31 @@ class Serializer {
...
+
+      // Accounts cannot have a tag.
+      if (decodedClassicAddress.tag) {
+        return undefined;
+      }
```
Stormtv: Somewhat similar here
```
if (decodedClassicAddress.tag !== undefined) return
```

src/serializer.ts
```diff
@@ -32,10 +32,31 @@ class Serializer {
...
+
+      normalizedAccount = decodedClassicAddress.address;
+    }
+    object["Account"] = normalizedAccount;
+    delete object.account;
+
     // Convert XRP denominated fee field.
     const txFee = transaction.getFee();
     if (txFee == undefined) {
```
Stormtv: Probably should use triple equals here

src/serializer.ts
keefertaylor: happy to convert to recommended style for the statement. 

I don't like omitting curly braces because I think it's error prone. Also assume you meant `return undefined` as opposed to `return`

src/serializer.ts
keefertaylor: Thank you, corrected throughout the file

PR #76, opened 2019-11-13
Merge latest protocol buffer updates

keefertaylor:
Specifically:
- Export request / response object for `getTransactionStatus` and `getLastValidatedLedger` RPCs
- Add support for `LastLedgerSequence` field of `Transaction` proto in `Serializer` class. 

I was able to use `npm link` to link these classes into AppianJS and build transactions that were successfully submitted to the ledger and either:
- failed to validate (`lastLedgerSequence` in the past)
- validated (`lastLedgerSequence` in the future)

### DIFF
--- a/test/serializer-test.ts
+++ b/test/serializer-test.ts
@@ -7,6 +7,7 @@ import { XRPAmount } from "../build/generated/xrp_amount_pb";
 import { assert } from "chai";
 import "mocha";
 import Utils from "../src/utils";
+const rippleCodec = require("ripple-binary-codec");
 
 /* eslint-disable @typescript-eslint/no-non-null-assertion */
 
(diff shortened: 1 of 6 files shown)


test/serializer-test.ts
```diff
@@ -16,9 +17,11 @@ describe("serializer", function(): void {
     const value = "1000";
     const destination = "rHb9CJAWyB4rj91VRWn96DkukG4bwdtyTh";
     const fee = "10";
+    const lastLedgerSequence = 20;
     const sequence = 1;
     const account = "r9LqNeG6qHxjeUocjvVki2XR35weJ9mZgQ";
-    const publicKey = "testPublicKey";
+    const publicKey =
```
keefertaylor: This is a real public key. I just updated them so that I could perform real data validation.

PR #88, opened 2019-12-02
Create a seed and derivation path initializer

keefertaylor:
Mnemonics and seeds are just different encodings for the same entropy buffer. Create a new initializer which allows users to create a HD wallet with a seed instead of forcing a mnemonic.

### DIFF
--- a/src/wallet.ts
+++ b/src/wallet.ts
@@ -83,7 +83,7 @@ class Wallet {
 
   /**
    * Generate a new hierarchical deterministic wallet from a seed and derivation path.
-   * 
+   *
    * @param {string} seed The seed for the wallet.
    * @param {string} derivationPath The derivation path to use. If undefined, the default path is used.
    * @returns {Wallet|undefined} A new wallet from the given mnemonic if the mnemonic was valid, otherwise undefined.
--- a/src/wallet.ts
+++ b/src/wallet.ts
@@ -78,6 +78,20 @@ class Wallet {
     }
 
     const seed = bip39.mnemonicToSeedSync(mnemonic);
+    return this.generateHDWalletFromSeed(seed, derivationPath);
+  }
+
+  /**
+   * Generate a new hierarchical deterministic wallet from a seed and derivation path.
+   * 
+   * @param {string} seed The seed for the wallet.
+   * @param {string} derivationPath The derivation path to use. If undefined, the default path is used.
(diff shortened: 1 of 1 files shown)


intelliot:
What are a few examples of XRP wallets and libraries that use a mnemonic? What’s the default derivation path?

keefertaylor:
In general, you'll need to be familiar with the following proposals:
[BIP32](https://github.com/bitcoin/bips/blob/master/bip-0032.mediawiki) - hierarchical generating seeds from seeds
[BIP39](https://github.com/bitcoin/bips/blob/master/bip-0039.mediawiki) - representing seeds as mnemonics
[BIP44](https://github.com/bitcoin/bips/blob/master/bip-0044.mediawiki) - standardizes a derivation path for HD wallets (builds on BIP32)
[SLIP44](https://github.com/satoshilabs/slips/blob/master/slip-0044.md) - Assigns values to currencies in the derivation path

----

> What are a few examples of XRP wallets and libraries that use a mnemonic? 
Xpring-JS, Xpring Wallet System, Ledger devices, any multi currency wallet application (Trust, Exodus, etc)

Note as well that we don't force users to use HD wallets, we still support seed only options: https://github.com/xpring-eng/xpring-common-js/blob/master/src/wallet.ts#L94.

> What’s the default derivation path?
This is the derivation path defined in BIP/SLIP44. It is made publicly available here: https://github.com/xpring-eng/xpring-common-js/blob/master/src/wallet.ts#L33.

src/wallet.ts
```diff
@@ -78,6 +78,20 @@ class Wallet {
...
     const seed = bip39.mnemonicToSeedSync(mnemonic);
+    return this.generateHDWalletFromSeed(seed, derivationPath);
+  }
+
+  /**
+   * Generate a new hierarchical deterministic wallet from a seed and derivation path.
+   *
+   * @param {string} seed The seed for the wallet.
+   * @param {string} derivationPath The derivation path to use. If undefined, the default path is used.
+   * @returns {Wallet|undefined} A new wallet from the given mnemonic if the mnemonic was valid, otherwise undefined.
```
Stormtv: Probably should be `given seed` etc...

keefertaylor: Done. Fixed comments elsewhere for consistency.

keefertaylor:
@intelliot I'm going to merge this PR. Happy to follow up on concerns you may have via GitHub issues or offline.

PR #104, opened 2019-12-16
Add `testnet` parameter when Encoding X-Addresses

keefertaylor:
X-Addresses support a special format for TestNet. Xpring-Common-JS currently ignores this parameter and defaults everything to mainnet. 

Add support for this parameter. Make the parameter optional and default to mainnnet. This is because I think:
- Most folks will just want mainnet addresses by default
- The behavior of existing code won't change, so we can release this as a point fix

Tested:
- Unit tests

Note that the unit test case is derived from manually checking values on http://xrpaddress.info.

### DIFF
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -70,11 +70,13 @@ class Utils {
    *
    * @param classicAddress A classic address to encode.
    * @param tag An optional tag to encode.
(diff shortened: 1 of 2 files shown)


keefertaylor:
Note that I renamed 'isTestNet' to 'test' which is what the proposal writers prefer: 
https://github.com/xrp-community/standards-drafts/issues/6

PR #117, opened 2019-12-26
Update Package.json

dangell7:
Updated `npm run webpack` script to add remote submodules before running  regenerate protos


dangell7:
My Error, See https://github.com/xpring-eng/xpring-common-js/blob/master/CONTRIBUTING.md

PR #122, opened 2020-01-06
Generate Protocol Buffers from rippled

keefertaylor:
Long term, [xpring-common-protocol-buffers](https://github.com/xpring-eng/xpring-common-protocol-buffers) is going to disappear and be replaced with an official and well reviewed implementation by the rippled team. 

The implementation in rippled is being done here: https://github.com/ripple/rippled/pull/3159.

To access the new rippled protocol buffers, a new submodule is added. Note that the submodule is `cjcobb23/rippled` (rather than `ripple/rippled`) since I can't figure out how to get a submodule to track an open PR. I filed an Asana task to track this update here: https://app.asana.com/0/1143722888069470/1155647690941937.

This change segregates the protocol buffers so that we can do the transition in several phases. As of now:
- /generated/*.proto : 'new' protocol buffers from rippled. Unused, as of now.
- /generated/legacy/*.proto : The same protocol buffers from xpring-common-protocol-buffers which we have been using. 

When we have fully migrated the library to rippled's protocol buffers we can simply delete the `legacy` folder to delete the migration. 

Note that this PR is basically the same as: https://github.com/xpring-eng/Xpring-JS/pull/108


scripts/regenerate_protos.sh
```diff
@@ -2,32 +2,69 @@
...
+# Generate typescript declaration files.
+$PWD/node_modules/grpc-tools/bin/protoc \
+    --plugin=protoc-gen-ts=./node_modules/.bin/protoc-gen-ts \
+    --ts_out=$LEGACY_TS_OUT_DIR \
+    --ts_out=$LEGACY_JS_OUT_DIR \
+    --proto_path=$PWD/xpring-common-protocol-buffers/proto \
+    $PWD/xpring-common-protocol-buffers/**/*.proto
 
-echo "All done!"
+echo "All done!"
```
hbergren: lint: newline

keefertaylor:
We don't really need to export them from here. The serializer takes a protocol buffer as an input and serializes it to a form the ripple libs can understand, so we should export that particular transaction. I will remove these in a follow up. 

Historically, I had done this because I was having some problems with typescript which turned out to be user error on my part.

PR #124, opened 2020-01-06
Mark existing Signing and Serialization functions as `legacy`

keefertaylor:

- [x] Rename existing functions to have a legacy prefix (this PR)
- [ ] Add new serialization functions and tests
- [ ] Add new signing functions and tests

Afterwards, client libraries will need to get upgraded. This will be a breaking change and should have a major version number bump when the next release is cut.


test/serializer-test.ts
```diff
@@ -1,5 +1,3 @@
-import { Currency } from "../generated/legacy/currency_pb";
-import { FiatAmount } from "../generated/legacy/fiat_amount_pb";
```
hbergren: 👍

keefertaylor:
> By "client libraries", do you mean the higher-level SDK libraries? Or users of the SDK?
Generally, the SDK libraries. Though anyone could bind up this library for their own use. 

> If it's a breaking change, it should have a major version number bump, not minor, right?
Assume we're talking <major>.<minor>.<fix>. I usually consider:
`major`: A major rearchitecture or feature addition
`minor`: A breaking change without a new features
`fix`: A minor fix that is not breaking that we'd like to just uptake

PR #125, opened 2020-01-06
Implement Serialization Methods and Tests for rippled protocol buffers

keefertaylor:

- [x] [Rename existing functions to have a legacy prefix](https://github.com/xpring-eng/xpring-common-js/pull/124)
- [x] Add new serialization functions and tests (this PR)
- [ ] Add new signing functions and tests

Afterwards, client libraries will need to get upgraded. This will be a breaking change and should have a minor version number bump when the next release is cut.


test/serializer-test.ts
```diff
@@ -263,4 +268,311 @@ describe("serializer", function(): void {
...
+    transaction.setPayment(payment);
+    transaction.setSigningPublicKey(Utils.toBytes(publicKey));
+    transaction.setLastLedgerSequence(lastLedgerSequence);
+
+    // WHEN the transaction is serialized to JSON.
+    const serialized = Serializer.transactionToJSON(transaction);
+
+    // THEN the result is as expected.
+    const expectedJSON = {
+      Account: Utils.decodeXAddress(account)!.address,
```
Stormtv: Nice usage of the non-null assertion 👍

src/serializer.ts
```diff
@@ -17,14 +19,156 @@ interface PaymentJSON {
...
 class Serializer {
+/**
+   * Convert a Transaction to a JSON representation.
+   *
+   * @param {proto.Transaction} transaction A Transaction to convert.
+   * @returns {Object} The Transaction as JSON.
+   */
+  public static transactionToJSON(
+    transaction: Transaction
+  ): object | undefined {
```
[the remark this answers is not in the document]
keefertaylor: Do you mean instead of `null`? I don't really have an opinion.

I tend to think of `undefined` as, the inputs were invalid and the output was not defined. 
Whereas null is: the inputs were valid but the output is nothing. 

I don't know if that's a great mental model. Happy to change if you and other reviewers feel strongly.

src/serializer.ts
```diff
@@ -17,14 +19,157 @@ interface PaymentJSON {
...
+    delete object.sourceTag;
+    delete object.accountTransactionId;
+
+    // Encode SigningPubKey to hex, which is what ripple-binary-codec expects.
+    object.SigningPubKey = Utils.toHex(transaction.getSigningPublicKey_asU8());
+
+    // Convert account field, handling X-Addresses if needed.
+    const accountAddress = transaction.getAccount();
+    if (!accountAddress) {
+      return undefined;
```
hbergren: You shouldn't need to `return undefined`. Just a `return` will do (unless Typescript or a linter is yelling about it).

src/serializer.ts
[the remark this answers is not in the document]
hbergren: My mental model for `undefined` and `null` in JS is:
- `undefined` means the JS engine did not assign a value to this particular variable
- `null` means there was a programmer that explicitly made a decision to clear out that variable.

I think it would be poor form to explicitly set a variable to `undefined`, but we have a function here.

For early returning from a function, `undefined` seems right. `void` is the type in Typescript that is commonly used for a function that returns nothing, so maybe that is more appropriate.

src/serializer.ts
keefertaylor: As a non-JavaScript programmer, I find this syntax really hard. `return` implies the function has no return type in most other languages.

I've made the change since I trust you to be more syntactically / idiomatic minded than myself.

src/serializer.ts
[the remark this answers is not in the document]
keefertaylor: > My mental model for undefined and null in JS is:
> - undefined means the JS engine did not assign a value to this particular variable
> - null means there was a programmer that explicitly made a decision to clear out that variable.

This matches mine, I think. IIUC, you're in favor of leaving this as is then?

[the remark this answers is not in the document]
hbergren: Yeah, I'm ok with leaving as is. I do think `return` is more idiomatic than `return undefined`, so thanks for that change.

src/serializer.ts
```diff
@@ -17,14 +19,157 @@ interface PaymentJSON {
...
+   * @returns {Object} The Transaction as JSON.
+   */
+  public static transactionToJSON(
+    transaction: Transaction
+  ): object | undefined {
+    // Serialize the protocol buffer to a JSON representation.
+    var object: any = transaction.toObject();
+
+    // Convert fields names where direct conversion is possible.
+    this.convertPropertyName("sequence", "Sequence", object);
```
hbergren: What is this method doing?

src/serializer.ts
```diff
@@ -17,14 +19,157 @@ interface PaymentJSON {
...
+      object
+    );
+
+    // Delete unsupported fields from the protocol buffer.
+    delete object.memosList;
+    delete object.flags;
+    delete object.signature;
+    delete object.signersList;
+    delete object.sourceTag;
+    delete object.accountTransactionId;
```
hbergren: You can do this in a single line:
https://stackoverflow.com/questions/32534602/javascript-built-in-function-to-delete-multiple-keys-in-an-object

Not sure if that's _better_. It might be less clear what's going on, but I tend to like succinct code.

src/serializer.ts
```diff
@@ -17,14 +19,157 @@ interface PaymentJSON {
...
+
+    // Encode SigningPubKey to hex, which is what ripple-binary-codec expects.
+    object.SigningPubKey = Utils.toHex(transaction.getSigningPublicKey_asU8());
+
+    // Convert account field, handling X-Addresses if needed.
+    const accountAddress = transaction.getAccount();
+    if (!accountAddress) {
+      return;
+    }
+    const account = accountAddress.getAddress();
```
hbergren: If `account` is an address, what is `accountAddress`?

src/serializer.ts
```diff
@@ -17,14 +19,157 @@ interface PaymentJSON {
...
+      }
+
+      normalizedAccount = decodedClassicAddress.address;
+    }
+    object["Account"] = normalizedAccount;
+    delete object.account;
+
+    // Convert XRP denominated fee field.
+    const txFee = transaction.getFee();
+    if (txFee === undefined) {
```
hbergren: I assume Typescript asserts it is either `undefined | number`?

test/serializer-test.ts
```diff
@@ -263,4 +268,311 @@ describe("serializer", function(): void {
...
+    const sender = new AccountAddress();
+    sender.setAddress(account);
+
+    const transaction = new Transaction();
+    transaction.setAccount(sender);
+    transaction.setFee(transactionFee);
+    transaction.setSequence(sequence);
+    transaction.setPayment(payment);
+    transaction.setSigningPublicKey(Utils.toBytes(publicKey));
+    transaction.setLastLedgerSequence(lastLedgerSequence);
```
hbergren: Do you want to move all these steps into a helper method that builds up a transaction? Seems like this structure is duplicated across tests.

test/serializer-test.ts
```diff
@@ -263,4 +268,311 @@ describe("serializer", function(): void {
...
+      Destination: "rU6K7V3Po4snVhBBaU29sesqs2qTQJWDw1",
+      Fee: fee.toString(),
+      LastLedgerSequence: lastLedgerSequence,
+      Sequence: sequence,
+      TransactionType: "Payment",
+      SigningPubKey: publicKey
+    };
+    assert.deepEqual(serialized, expectedJSON);
+  });
+});
```
hbergren: nit: newline

src/serializer.ts
keefertaylor: `AccountAddress` is a special protocol buffer that wraps addresses. This gives us slightly better typing. Imagine:

```
message AccountInfo {
   string address = 1;
}
```

vs

```
message AccountAddress {
  string address = 1;
}

message AccountInfo {
  AccountAddress address = 1;
}
```

In the former case, you can stuff anything you want in the `string` field, where in the later, we've used a wrapper type to make it clear what specific KIND of string should be placed in it.

src/serializer.ts
keefertaylor: Correct.

test/serializer-test.ts
keefertaylor: Done. I also tried to pull out certain fields to constants as well. This file is getting unwieldy at the moment. Fortunately, we should be able to deprecate and remove the `legacy` code soon which should make things a bit better.

src/serializer.ts
keefertaylor: This method re-writes the names of fields on an object. The format that rippled expects isn't the widely accepted JSON that the protocol buffer formats itself to. 

Deferring the conversation on whether this is a good idea to the main comment thread.

src/serializer.ts
keefertaylor: Sure. I think that this obfuscates the code meaning a bit, but it's not too complex so I'm fine with tightening up the code a bit.

keefertaylor:
> Does it make sense for the serializer to build up a separate final object instead of mutating the object we get from transaction.toObject?

Ultimately, I had hoped that this class could be: 
```
const jsonTransaction = transaction.toObject();
const hexTransaction = ripple-binary-codec.encode(jsonTransaction)
return hexTransaction;
```

The idea being that we can utilize more and more of the autogenerated protocol buffer code in order to do this and rely less and less on hand rolled logic.  I don't think this is going to make the flow easier to follow, but I'm happy to do it if you think there's a need. I think that every `rewrite` is just replaced with a hardcoded assignment, and the conversion code is largely the same. At first I thought that the rewrite might make the conversion methods more functional, but they're already stateless and I don't think it makes a difference. 

I also note that I'm being lazy, so I'm happy to hear the case you'd like to make on how this is is clearer, then I'm inclined to do it.

(I really dream that ripple-binary-codec actually works with protocol buffers and this class disappears, but I think that's premature and a fair amount of work).

test/serializer-test.ts
```diff
@@ -263,4 +232,112 @@ describe("serializer", function(): void {
...
+      Amount: value.toString(),
+      Destination: destinationClassicAddress,
+      Fee: fee.toString(),
+      LastLedgerSequence: lastLedgerSequence,
+      Sequence: sequence,
+      TransactionType: "Payment",
+      SigningPubKey: publicKey
+    };
+    assert.deepEqual(serialized, expectedJSON);
+  });
```
hbergren: Great job adding more tests!

keefertaylor:
Makes sense to me. I'll try to break off some of these methods and send you some PRs to clean it up over time. 

I'd really like to merge this because it unblocks upstream dependencies, so I'm going to do that now.

PR #126, opened 2020-01-06
Implement Signing Methods and Tests for rippled protocol buffers 

keefertaylor:

- [x] [Rename existing functions to have a legacy prefix](https://github.com/xpring-eng/xpring-common-js/pull/124)
- [x] [Add new serialization functions and tests](https://github.com/xpring-eng/xpring-common-js/pull/125)
- [x] Add new signing functions and tests

Afterwards, client libraries will need to get upgraded. This will be a breaking change and should have a minor version number bump when the next release is cut.

### DIFF
--- a/src/signer.ts
+++ b/src/signer.ts
@@ -33,8 +33,7 @@ class Signer {
     }
     const transactionHex = rippleCodec.encodeForSigning(transactionJSON);
 
-    const signatureHex = wallet.sign(transactionHex);
-    return signatureHex;
+    return wallet.sign(transactionHex);
   }
 
   /**
--- a/src/signer.ts
+++ b/src/signer.ts
@@ -2,7 +2,8 @@
 
 import Serializer from "./serializer";
 import { SignedTransaction } from "../generated/legacy/signed_transaction_pb";
-import { Transaction } from "../generated/legacy/transaction_pb";
+import { Transaction as LegacyTransaction } from "../generated/legacy/transaction_pb";
+import { Transaction } from "../generated/rpc/v1/transaction_pb";
 import Wallet from "./wallet";
 
 const rippleCodec = require("ripple-binary-codec");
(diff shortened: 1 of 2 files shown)


src/signer.ts
```diff
@@ -11,6 +12,31 @@ const rippleCodec = require("ripple-binary-codec");
...
+    }
+
+    var transactionJSON = Serializer.transactionToJSON(transaction);
+    if (transactionJSON === undefined) {
+      return undefined;
+    }
+    const transactionHex = rippleCodec.encodeForSigning(transactionJSON);
+
+    const signatureHex = wallet.sign(transactionHex);
+    return signatureHex;
```
hbergren: Do you want the `signatureHex` variable here to name this, rather than just returning it?

test/signer-test.ts
```diff
@@ -55,4 +57,53 @@ describe("signer", function(): void {
...
+    transaction.setFee(transactionFee);
+    transaction.setSequence(sequence);
+    transaction.setPayment(payment);
+
+    // WHEN the transaction is signed with the wallet.
+    const signature = Signer.signTransaction(transaction, wallet);
+
+    // THEN the signing artifacts are as expected.
+    assert.exists(signature);
+    assert.isTrue(Utils.isHex(signature!));
```
hbergren: Neat. TIL about `variable!` in Typescript.

src/signer.ts
keefertaylor: Nope, just a poorly finished thought after copy / pasting the legacy method below.

keefertaylor:
> approved, I do prefer returning null vs undefined though.

There was some talk of this in #125. Happy to hear why you think one is definitely better than the other: https://github.com/xpring-eng/xpring-common-js/pull/125#discussion_r364383569.

PR #141, opened 2020-01-18
Basic gRPC-Web (1 of 4)

Stormtv:
The goal of these changes is to reduce the number of libraries we have to maintain such as Xpring-Web and Xpring-JS. The common goal is to have a browser and node.js compatible library.

### Context of Change
This first part of the 4 part PR is to swap out the node.js gRPC to gRPC-web. In addition the generated files are now included under /src/ so there are some path changes as well.

### Type of Change

## Before / After
gRPC web is now used rather than gRPC node

## Test Plan
Passes CI


src/serializer.ts
```diff
@@ -56,7 +56,7 @@ class Serializer {
       return;
     }
 
-    var normalizedAccount = account;
+    let normalizedAccount = account;
```
keefertaylor: Why do you prefer `let` over `const` or `var` here?

keefertaylor:
One question, but this looks reasonable to me. Please wait for a second reviewer to approve.

src/serializer.ts
Stormtv: Best practice is to not use var anymore as let is a more strict option since it is block scoped instead of function scoped. Let is best for variables that change value like this one (line 71) where as const is best for variables that do not change value.

hbergren: When he says "`let` is best for variables that change value." he means, "You couldn't change the value of `normalizedAccount` like you do on line 71 if it was `const`."

keefertaylor:
I patched this into XpringKit and I am happy.

scripts/regenerate_protos.sh
```diff
@@ -9,32 +9,20 @@ set -e -o pipefail
...
+mkdir -p $OUT_DIR
 
 # Generate node code.
 $PWD/node_modules/grpc-tools/bin/protoc \
-    --js_out=import_style=commonjs,binary:$TS_OUT_DIR \
-    --js_out=import_style=commonjs,binary:$JS_OUT_DIR \
-    --grpc_out=$TS_OUT_DIR \
-    --grpc_out=$JS_OUT_DIR \
+    --js_out=import_style=commonjs,binary:$OUT_DIR \
+    --grpc-web_out=import_style=commonjs+dts,mode=grpcwebtext:$OUT_DIR \
```
hbergren: Does the `dts` here mean we still ship typings?

tsconfig.json
```diff
@@ -11,10 +11,12 @@
     "strict": true,
     "resolveJsonModule": true,
     "esModuleInterop": true,
+    "allowJs": true,
+    "declaration": true,
```
hbergren: Do we need `declarationMap` as well? I always get a little confused on TypeScript compiler settings.

(If we don't, please explain to me why not).

tsconfig.json
```diff
@@ -11,10 +11,12 @@
...
     "typeRoots": ["./src/types", "./node_modules/@types"]
   },
   "include": [
     "src/**/*",
-    "generated/*"
+    "test/**/*"
```
hbergren: Since we now include `test` in our `tsconfig`, see if you can get rid of `tsconfig.eslint.json` and change the `project` key in `.eslintrc.js` back to the standard `tsconfig.json` and still lint successfully.

tsconfig.json
Stormtv: declarationMap just adds the support for mapping declaration files to their original source. When distributing the build we don't include the original `src` directory so having the maps here would be pointless unless we included the original source files in the build.

tsconfig.json
Stormtv: Weird things happened with the rebase I'll remove that.

scripts/regenerate_protos.sh
Stormtv: Yes the protos will produce a commonjs js file and a typescript typings file along with it.

PR #142, opened 2020-01-18
Update TSConfig (2 of 4)

Stormtv:
Part two updates the TSConfig. Remove tests from compiler options and makes it so we don't generate /src/ or /test/ directories anymore inside our build. Also we default to es5 out which is a cleaner option to allow for browsers to consume even without webpack.

### Context of Change
This refactor makes the build output cleaner.

### Type of Change

## Before / After
Build had /src/ directory and /test/ directory now it is flattened.

## Test Plan
CI Passes


package.json
```diff
@@ -47,6 +47,7 @@
     "nyc": "15.0.0",
     "prettier": "^1.19.1",
     "source-map-support": "0.5.16",
+    "ts-loader": "^6.2.1",
```
hbergren: Why did we add `ts-loader` as a dependency?

tsconfig.json
```diff
@@ -1,22 +1,27 @@
...
-    "esModuleInterop": true,
-    "allowJs": true,
-    "declaration": true,
+    "isolatedModules": true,
+    "sourceMap": true,
     "typeRoots": ["./src/types", "./node_modules/@types"]
   },
   "include": [
-    "src/**/*",
-    "test/**/*"
```
hbergren: Why did we removethe test directory?

tsconfig.json
```diff
@@ -1,22 +1,27 @@
...
     "module": "commonjs",
     "moduleResolution": "node",
-    "sourceMap": true,
-    "removeComments": true,
-    "strict": true,
     "resolveJsonModule": true,
-    "esModuleInterop": true,
-    "allowJs": true,
-    "declaration": true,
+    "isolatedModules": true,
```
hbergren: Why did you enable `isolatedModules`?

tsconfig.json
```diff
@@ -1,22 +1,27 @@
 {
   "compilerOptions": {
-    "baseUrl": "src",
-    "rootDir": "./",
     "outDir": "build",
-    "target": "esnext",
+    "target": "es5",
```
hbergren: ES2017 is supported by all major browsers. I don't think you need to go as low as ES5.

tsconfig.json
```diff
@@ -1,22 +1,27 @@
...
+    "lib": [
+      "dom",
+      "dom.iterable",
+      "esnext"
```
hbergren: Why do you have `esnext` in the `lib` section?

tsconfig.json
```diff
@@ -1,22 +1,27 @@
...
+    ],
+    "allowJs": true,
+    "skipLibCheck": true,
+    "declaration": true,
+    "esModuleInterop": true,
+    "allowSyntheticDefaultImports": true,
```
hbergren: `allowSyntheticDefaultImports` is true by default with `esModuleInterop`, so you don't need this.

tsconfig.json
```diff
@@ -1,22 +1,27 @@
...
-    "rootDir": "./",
     "outDir": "build",
-    "target": "esnext",
+    "target": "es5",
+    "lib": [
+      "dom",
+      "dom.iterable",
+      "esnext"
+    ],
+    "allowJs": true,
```
hbergren: Why do we have `allowJs` enabled?

keefertaylor:
I am working on patching this PR into XpringKit and verifying the functionality still works.

tsconfig.json
Stormtv: It perform additional checks to ensure that separate compilation such as with babel would be safe. I assume a common way of digesting xpring-common-js ,the way Xpring Wallet will be using it, will be through babel.

tsconfig.json
Stormtv: There is no need to have it in the includes in the tsconfig.json compile options it is linted through the tsconfig.eslint.json

package.json
Stormtv: This should of been included on the next PR step, removing from here.

tsconfig.json
Stormtv: Alright I'll change this to ES2017

tsconfig.json
Stormtv: Nice, removed this

tsconfig.json
Stormtv: The output of the protos are .js files along with typing files so in order to consume them we need to enable allowJs

tsconfig.json
Stormtv: I guess we might not need it in this lib we could go down a version but I did "esnext" for since we were using BigInt in Xpring-JS. I'm not sure what the harm is in using esnext it doesn't impact build size just what we are allowed to use in the source code.

Stormtv:
@keefertaylor I had to comment out the Serializer.convertPropertyName as it was triggering the noUnusedLocals. Is this function able to be removed or what would you prefer to be done here?

src/serializer.ts
```diff
@@ -287,18 +287,18 @@ class Serializer {
...
-    newPropertyName: string,
-    object: any,
-    /* eslint-enable @typescript-eslint/no-explicit-any */
-  ): void {
-    /* eslint-disable no-param-reassign */
-    object[newPropertyName] = object[oldPropertyName]
-    delete object[oldPropertyName]
-    /* eslint-enable no-param-reassign */
-  }
+  // private static convertPropertyName(
```
keefertaylor: This should be fixed in #152

src/index.ts
```diff
@@ -12,12 +12,10 @@ export { SubmitSignedTransactionRequest } from './generated/legacy/submit_signed
...
 export { SignedTransaction } from './generated/legacy/signed_transaction_pb'
-export { default as Signer } from './signer'
 export { TransactionStatus } from './generated/legacy/transaction_status_pb'
 export { Transaction } from './generated/legacy/transaction_pb'
 
 export { default as Wallet } from './wallet'
-export { WalletGenerationResult } from './wallet'
 export { default as Utils } from './utils'
-export { ClassicAddress } from './utils'
+export { default as Signer } from './signer'
```
keefertaylor: Please export ClassicAddress

src/serializer.ts
Stormtv: pulled in

Stormtv:
@keefertaylor 
> Can you please patch this into XpringKit or Xpring4J?
> You should only need to do this in one of them since they both behave the same, so if it passes for one it will pass for both.

Pulled it in and this PR version maintains compatibility with XpringKit

Screen Shot 2020-01-23 at 2 24 15 PM

src/index.ts
Stormtv: Okay

tsconfig.json
Stormtv: Removed this for now

There is some additional information on babel about this if you are interested:
https://babeljs.io/docs/en/babel-plugin-transform-typescript#typescript-compiler-options
https://babeljs.io/docs/en/babel-plugin-transform-typescript

src/index.ts
```diff
@@ -12,12 +12,14 @@ export { SubmitSignedTransactionRequest } from './generated/legacy/submit_signed
...
 export { default as Serializer } from './serializer'
+
+// Type Exports
```
hbergren: 😄 Good comment. Thanks for giving context.

tsconfig.json
hbergren: Sure, this is probably fine.

I'm unhappy not because of the code you wrote, but because this is the least well documented part of the TypeScript compiler settings. It's undocumented what `libs` are included by default when your target is `es2017`, the use case for `lib` is pretty much undocumented... I just don't like configuration settings I don't understand well.

But we are using `BigInt`. That does mean that if `xpring-common-js` _is_ used in the browser, they may need to polyfill `BigInt`: https://caniuse.com/#search=bigint

tsconfig.json
```diff
@@ -1,18 +1,37 @@
...
+    "noImplicitReturns": true,
+    "noFallthroughCasesInSwitch": true,
+
+    // Module Resolution Options
+    "moduleResolution": "node",
+
+    // Advanced Options
     "allowJs": true,
-    "declaration": true,
+    "skipLibCheck": true,
```
hbergren: Why do we `skipLibCheck`?

Does anything break if we remove this line?

Stormtv: I added this back when we were doing the weird requires thing with the any's I removed this now good catch.

tsconfig.json
Stormtv: If we omitted the lib entry the default lib for target of ES2017 would be `DOM,ES2017,DOM.Iterable`

When working with Lib's I refer to the source code to see what each version gives us. 
https://github.com/Microsoft/TypeScript/tree/master/lib
*esnext is es2020*

Yes, we do need to polyfill bigInt this is one of the PR's I will have out today actually. Once this is polyfilled we should probably just use the default lib options for the ES2017 so that we will see errors when using things that should probably be polyfilled.

BigInt is only used in Xpring-JS so I can actually make that change now on this Library.

PR #143, opened 2020-01-18
Update Webpack (3 of 4)

Stormtv:
Produce a minified cross platform compatible UMD webpack output. As well as refactor out the deprecated mocha.opts.

We now utilize ts-loader so we don't have to run the typescript compiler and then webpack the output now we just run webpack and it does both.

### Context of Change
Refactors the TSConfig to produce a UMD that is consumable by both node.js and browsers when webpacked.

Fix deprecated use of Mocha.opts

Restructures the scripts to lint and test on build

### Type of Change

## Before / After
Removes the deprecation error with mocha, and produces a umd webpack output.

## Test Plan
CI continues to pass


package.json
```diff
@@ -19,13 +19,14 @@
...
   },
   "scripts": {
-    "build": "tsc -d",
-    "clean": "rm -rf ./generated ./dist ./build",
-    "pretest": "./scripts/regenerate_protos.sh && tsc",
-    "lint": "eslint . --ext .ts --fix",
-    "test": "npm run lint && nyc mocha",
-    "prepublishOnly": "npm test && tsc -d",
-    "webpack": "npm run clean && ./scripts/regenerate_protos.sh && tsc && npx webpack"
+    "build": "rm -rf ./src/generated ./build && ./scripts/regenerate_protos.sh && npm run lint && tsc -d",
```
keefertaylor: Can you update this for @hbergren's build changes? Specifically I think he removed `lint:fix`.

Also think that you should just call `npm run clean` as the first step of thebuild command.

Stormtv: Fixed, I do think we should have kept fix and lint as two separate commands though.

Stormtv:
> Can you confirm this also builds in XpringKit or Xpring4J?

It works with this compatibility change https://github.com/xpring-eng/XpringKit/pull/83
Java fix PR will come soon.

There are two changes that are breaking compatability:
EntryPoint renamed to XpringCommonJS
bundled.js renamed to index.js

The reason for changing the entry point is to allow other consumers of the webpack to have a unique and meaningful variable name when using the script in a browser. This isn’t much of an issue for XpringKit or XpringJ4.

The reason for chacing the bundled to index is so the typescript definitions work for browser clients when using the webpack build. We could use a shell command to rename this file to bundled but index makes more sense when requiring the library into the browser also.
Overall I think these are good changes and only 6 lines of code need to be changed inside XpringKit. Although we could lose the changes that I mentioned above and maintain compatibility at the cost of a worse developer experience when working with the webpack distribution.

package.json
hbergren: Why do you prefer separate steps for linting and fixing?

My thinking is that anything that can be automatically fixed just should be, without you thinking about it. `lint` is the command usually run in CI, but I don't feel like CI should fail because someone forgot to run Prettier before pushing.

package.json
```diff
@@ -19,13 +19,13 @@
     "grpc-web": "1.0.7"
   },
   "scripts": {
-    "build": "tsc -d",
-    "clean": "rm -rf ./generated ./dist ./build",
-    "pretest": "./scripts/regenerate_protos.sh && tsc",
+    "build": "npm run clean && ./scripts/regenerate_protos.sh && npm run lint && tsc -d",
+    "webpack": "npm run clean && ./scripts/regenerate_protos.sh && npm run lint && npx webpack",
+    "clean": "rm -rf ./src/generated ./dist ./build",
+    "pretest": "npm run clean && ./scripts/regenerate_protos.sh && npm run lint",
```
hbergren: Nice moving linting into `pretest`. That's a cleaner way to think about it.

package.json
```diff
@@ -63,5 +64,14 @@
     ],
     "all": true,
     "check-leaks": true
+  },
+  "mocha": {
```
hbergren: Can you move mocha configuration into a `.mocharc.js` file?

Here's the example Mocha config JS file:
https://github.com/mochajs/mocha/blob/master/example/config/.mocharc.js

And here's the example YML file with every configuration option:
https://github.com/mochajs/mocha/blob/master/example/config/.mocharc.yml

webpack.config.js
```diff
@@ -1,8 +1,32 @@
+// eslint-disable-next-line @typescript-eslint/no-var-requires
+const path = require('path')
+
 module.exports = {
-  entry: './build/index.js',
+  target: 'web',
+  mode: 'production',
+  entry: './src/index.ts',
+  devtool: 'false',
```
hbergren: Why no sourcemaps?

package.json
```diff
@@ -19,13 +19,13 @@
     "grpc-web": "1.0.7"
   },
   "scripts": {
-    "build": "tsc -d",
-    "clean": "rm -rf ./generated ./dist ./build",
-    "pretest": "./scripts/regenerate_protos.sh && tsc",
+    "build": "npm run clean && ./scripts/regenerate_protos.sh && npm run lint && tsc -d",
+    "webpack": "npm run clean && ./scripts/regenerate_protos.sh && npm run lint && npx webpack",
```
hbergren: I get that now we have `ts-loader` we don't _need_ to run `tsc`, but wouldn't it be nice to make sure our TypeScript compiles without errors?

Maybe we can have a command that calls both `build` and `webpack`? (Might need to rename some of the commands).

What do you think? Or am I being overly cautious?

webpack.config.js
```diff
@@ -1,8 +1,32 @@
...
+  module: {
+    rules: [
+      {
+        test: /\.tsx?$/,
```
hbergren: nit: We won't have any `.tsx` files in here

keefertaylor:
> There are two changes that are breaking compatibility:

These make sense. As long as the functionality is there, I don't mind doing shim updates.

webpack.config.js
Stormtv: Good point, added that in.

package.json
Stormtv: I think it is overly cautious as each CI has to pass all compilations.

I agree that this would be nice to have nice to have and doesn't increase our build times significantly so I added in a tsc --noEmit to the webpack script and the test script.

webpack.config.js
Stormtv: Okay removed the optional x

.mocharc.js
```diff
@@ -0,0 +1,11 @@
...
+
+module.exports = {
+  require: [
+    'ts-node/register',
+    'source-map-support/register',
+  ],
+  extension: [
+    'ts'
+  ],
+}
```
hbergren: nit: newline

.mocharc.js
```diff
@@ -0,0 +1,11 @@
+'use strict';
+
+module.exports = {
+  require: [
```
hbergren: Can we add a few more settings to the `mocha` configuration?

```
  opts: false,
  slow: 500,
  timeout: 5000,

  // Check for global variable leaks
  "check-leaks": true,
```

`opts:false` tells Mocha not to look for the `mocha.opts` file.
`slow` and `timeout` both warn if tests take too long. I'm not sure what good values for those are for this library, but these are the values used in `ripple-lib`. And then checking for global variable leaks is just a good thing to do I feel like.

PR #144, opened 2020-01-18
Update Version and Changelog (4 of 4)

Stormtv:
Adds a change log and increments the version to 2.1.0

### Context of Change
Documentation and versioning change

### Type of Change

## Before / After
Version goes up, changelog changes

## Test Plan
CI Passes - In addition to this being the 4 of 4 I would also test with the outstanding compatibility PRs found on Xpring-JS


CHANGELOG.md
```diff
@@ -7,6 +7,18 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+## [2.1.0] - 2020-01-TBD
+
+### Added
+
+The release switches us over to using gRPC-Web instead of the node.js version of gRPC. The goal of this change is to enabled the library to be used in browser as well as on Node.js without having two separate codebases or build strategies.
```
tedkalaw: nit: consistent capitalization for Node.js

CHANGELOG.md
```diff
@@ -7,6 +7,18 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+
+- Move from gRPC node to gRpc Web
```
tedkalaw: gRPC-Web

CHANGELOG.md
```diff
@@ -7,6 +7,18 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+- Update ESLint and configure it to work with prettier and TSConfig.
+- Fixed the lint issues found in the src as well as the tests.
+- Moved the generated files that are being consumed by the build process into the src.
+- Updated our webpack configuration to use UMD (Universal module definitions) to support both web and node.js.
```
tedkalaw: the last change uses the present tense to describe changes instead of past tense; would be good to stay consistent here

keefertaylor:
Please make sure we don't lose the other pending update on this PR

CHANGELOG.md
keefertaylor: "us over to using" -> "this library to using"

Generally prefer to refer to things objectively rather than the collective 'we'. There's probably some named grammar rule for this.

keefertaylor: s/enabled/enable

CHANGELOG.md
```diff
@@ -10,6 +10,15 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Breaking Changes
 - Modify `Signer`'s `signTransaction` method to return an array of signed bytes rather than a string based signature.
```
hbergren: @keefertaylor , this is a breaking change for consumers of `xpring-common-js`, right?

If so, since we're following SemVer, we should probably version to `3.0.0`, since we have a breaking, non backwards compatible change.

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "xpring-common-js",
-  "version": "2.0.1",
+  "version": "2.1.0",
```
hbergren: Since we increment the version number here, we should add a new `## [2.1.0]` header in the CHANGELOG.

(Or potentially a `## [3.0.0]` header depending on what Keefer says about the breaking change).

CHANGELOG.md
```diff
@@ -10,6 +10,15 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
 
+### Added gRPC-Web
+The release switches the library over to using gRPC-Web instead of the Node.js version of gRPC. The goal of this change is to enable the library to be usable on the web and Node.js without having two separate codebases or build strategies.
+
+- Moved from gRPC node to gRPC-Web
+- Updated ESLint and configured it to work with prettier and TSConfig.
+- Fixed the lint issues found in the src as well as the tests.
+- Moved the generated files that are being consumed by the build process into the src.
+- Updated webpack configuration to use UMD (Universal module definitions) to support both web and Node.js.
```
hbergren: I think these can be broken up a little more.

Some of these are only internal, like updating ESLint and fixing lint issues, so could go under an `### Internal` or `### Other` heading.

I'm not sure if moving the generated files is internal or not, so not sure where to put that.

I'm also not sure I would put `gRPC-Web` directly into the `### Added` header.

This is the changelog format we're trying to be based off: https://keepachangelog.com/en/1.0.0/

I also really like the Mocha changelog (although it deviates from that format a little):
https://github.com/mochajs/mocha/blob/master/CHANGELOG.md

Stormtv:
I removed the version change as there will have to be another PR that updates version and moves the content from unreleased to a proper version number.

CHANGELOG.md
```diff
@@ -9,6 +9,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Breaking Changes
 - Modify `Signer`'s `signTransaction` method to return an array of signed bytes rather than a string based signature.
+- Switched the library over to using gRPC-Web instead of the Node.js version of gRPC. The goal of this change is to enable the library to be usable on the web and Node.js without having two separate codebases or build strategies
```
hbergren: Is switching to `grpc-web` a breaking change for a user of `xpring-common-js`?

If so, why?

Stormtv: Technically it is not a breaking change for Xpring-Common-JS so I moved it to the added column. Sorry for the confusion.

PR #145, opened 2020-01-18
Refactor Serializer to Build a New Object

keefertaylor:

Serializer's implementation changes a protocol buffer to JSON. Then the JSON is mangled to make a JSON object that is compatible with ripple-binary-codec. 

This change is a refactor to build a new object, rather than mangling the JSON output of a protocol buffer. 

### Context of Change

This refactor makes the code less brittle and easier to read.  These changes were suggested in PR: https://github.com/xpring-eng/xpring-common-js/pull/125.

I did not modify the legacy code as it is deprecated and will be deleted eventually. 

### Type of Change

## Before / After

N/A

## Test Plan

Continuous integration tests.


src/serializer.ts
```diff
@@ -29,22 +29,11 @@ class Serializer {
...
-    this.convertPropertyName("signingPublicKey", "SigningPubKey", object);
-    this.convertPropertyName(
-      "lastLedgerSequence",
-      "LastLedgerSequence",
-      object
-    );
-
-    // Delete unsupported fields from the protocol buffer.
-    ['memosList', 'flags', 'signature', 'signersList', 'sourceTag', 'accountTransactionId'].forEach((key): boolean => delete object[key]);
+    var object: any = {}
```
Stormtv: Replace `var` and could we type this object now?

keefertaylor: Done.

keefertaylor:
Pushed some additional lint fixes into this PR (remove use of `any`, add `default` cases to switch statements).

PR #148, opened 2020-01-20
Encode signature into transaction bytes

keefertaylor:

Modify an the sign API in `Signer` to return an array of bytes representing both a signed transaction and a signature, rather than a hex string containing a signature. 

This unblocks work to migrate to the new rippled protocol buffers.

Xpring-Common-JS needs to be able to take a `Transaction` protocol buffer and produce a string from it. 
### Context of Change

rippled is publishing a built in gRPC layer. This library began to develop gRPC bindings before the code was fully merged. The code to publish protocol buffers was added in this PR: https://github.com/xpring-eng/xpring-common-js/pull/122. This change is required to publish a set of bytes in the new submit transaction protocol buffers. 

Previously, I used a (somewhat silly) system where I sent the transaction signature as a detached field from the transaction itself. This system had it's own set of problems, and we shouldn't use it. 

The [new gRPC interface in rippled wants only bytes](https://github.com/cjcobb23/rippled/blob/7d867b806d70fc41fb45e3e61b719397033b272c/src/ripple/proto/rpc/v1/submit.proto#L10).  It appears that the `TxnSignature` field is not deterministically located in the transaction (See #2 here: https://xrpl.org/serialization.html). In practice, this means that I have to encode a protocol buffer to bytes, sign the bytes, then re-encode the protocol buffer again with the signature. 

To do this, I've implemented the following:
- Convert `Transaction` protocol buffer to a JSON object (in `Serializer`). 
- Encode the JSON object to a hex string, by using ripple-binary codec's `encodeForSigningMethod`
- Generate a hex encoded string representing the signature, by using `Wallet`'s `sign` method
- Re-encode the `Transaction` protocol buffer to a JSON object, this time including the signature (using `Serializer`)
- Convert the JSON object to a hex encoded string by using ripple-binary-codec's encode method. 

### Type of Change

## Before / After

After this change, clients will be able to produce the inputs they need to use the new protocol buffers. 

## Test Plan

Added new unit tests and they run / pass on our continuous integration build.


keefertaylor:
> What makes us think that ripple-lib is too large of a dependency?
ripple-lib has a bunch of functionality and I'm not sure there's an API that would work with this project out of the box. I'd be fine adding a dependency, but it seems silly to do just for this one piece of functionality. 

> I'm neutral about Uint8Array vs. hex encoded strings. 

I prefer bytes so I've made this change. `Utils.toHex` is easy to invoke to change into a `string` if needed for UI display purposes or for debugging.

keefertaylor:
PTAL - I'm now happy for this to be reviewed.

PR #149, opened 2020-01-21
ESLint + Prettier

hbergren:

Add the standardized JS/TS linting configuration to `xpring-common-js`.

### Context of Change

The goal here is to get all of our JS/TS projects on the same configuration so we have some consistency across projects. This linting configuration is live on `ripple-keypairs` and will be migrated to the other `ripple-lib` style libraries moving forward.

Next steps after this PR are `xpring-js` and then perhaps the wallet (with some tweaks for Jest + React).

### Note for Reviewers

I tried to make my commits fairly isolated, so it is useful to review commit by commit as opposed to all at once.

I manually verified every change that Prettier made to ensure that no unexpected functionality was introduced. (Although the more I use Prettier, the more faith I have in it).

### Type of Change

### TODO

This won't pass CI right now because there are still 3 linting errors (and 7 warnings, but CI can pass with warnings).

The 3 errors are all about `case-switch` statements missing a `default` case in `serializer.ts`. I'm going to defer to @keefertaylor on what the right thing to do for that is.

Maybe it's impossible to reach the default case? If so, should we throw an error on default?

The 7 warnings are:
- We use `any` in a few places, and we shouldn't.
- We use `var x = require('x')` instead of `import * as x from 'x'` in a few places. This is a valid warning, because `require()` syntax doesn't give us type information for dependencies, but `ripple-binary-codec` is still written in JS, and `ripple-keypairs` does not (yet) ship type definitions, so I switched this rule from `error` to `warning` for the moment.

Ideally I'd like to have 0 errors and 0 warnings so the mental diff is higher impact if a linting regression is introduced, but I'm not sure if silencing those two warnings is the right thing to do.

## Before / After

Now we have a more robust linting configuration, and Prettier for auto-formatting. Some linting errors were also fixed.

## Test Plan

Run the tests, they all still pass.


hbergren:
Just pulled in Tyler's fixes for the require linting issue.

I am planning on cutting a release of `ripple-keypairs` this week that ships type definition files, so when that lands we can turn around and remove the empty module declaration file here.

package.json
```diff
@@ -25,29 +25,29 @@
...
     "test": "npm run lint && nyc mocha",
     "prepublishOnly": "npm test && tsc -d",
     "webpack": "npm run clean && ./scripts/regenerate_protos.sh && tsc && npx webpack"
   },
   "devDependencies": {
     "@types/chai": "4.2.7",
     "@types/mocha": "5.2.7",
-    "@typescript-eslint/eslint-plugin": "1.13.0",
-    "@typescript-eslint/parser": "1.13.0",
+    "@typescript-eslint/eslint-plugin": "^2.16.0",
```
keefertaylor: In general, I'm OCD so I prefer version locking. Any reason you're unlocking these versions?

keefertaylor:
This is great. Thanks for setting it up. 

We can add default cases. I can take care of doing that.

package.json
hbergren: I like getting new functionality without having to think about it.

I feel you on version locking for core/runtime dependencies, but I don't feel it's necessary for our linters. I'm happy to version lock if that is a strong preference.

keefertaylor: Nope that's a fine way to look at it. Carry on.

keefertaylor:
We're going to merge this (and knowingly break the build) so that @Stormtv is unblocked.

PR #154, opened 2020-01-24
fix XpringJS / Xpring4J

Stormtv:
fixes mistake that broke XpringJS / Xpring4J (this shouldn't have been included in part 1 of the gRPC web fix)

### Context of Change
rename the output back to bundled.js

### Type of Change
- [X] Bug fix (non-breaking change which fixes an issue)

## Test Plan
Pull into XpringKit and run tests

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -1,7 +1,7 @@
 module.exports = {
   entry: './build/src/index.js',
   output: {
-    filename: 'index.js',
+    filename: 'bundled.js',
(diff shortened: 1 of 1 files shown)


Stormtv:
https://github.com/xpring-eng/xpring-common-js/pull/142 fixes this also so if that goes in first we can close this.

keefertaylor:
We can just wait on #142, which looks like you're just waiting on answering a last question on.

PR #157, opened 2020-01-27
Fix build generated typings

Stormtv:
Not all typings were not properly built in the v3.0.0 update. This PR fixes that.

### Context of Change
Copies the generated typings from the source to the build / dist after build or webpack.
Also removes the `grpc_tools_node_protoc_ts` as it doesn't do anything when using gRPC-Web with commonjs+dts

### Type of Change
- [X] Bug fix (non-breaking change which fixes an issue)

## Before / After
The typings found in the generated directory under build and dist are now correct instead of empty exports.

## Test Plan
CI passes and code can be imported into Xpring-JS without the skipLibCheck tsConfig option.

### DIFF
--- a/package.json
+++ b/package.json
@@ -19,8 +19,8 @@
     "ripple-keypairs": "^0.11.0"
   },
   "scripts": {
-    "build": "npm run clean && ./scripts/regenerate_protos.sh && npm run lint && tsc -d && rsync -r ./src/generated ./build",
-    "webpack": "npm run clean && ./scripts/regenerate_protos.sh && npm run lint && tsc --noEmit && npx webpack && rsync -r ./src/generated ./dist",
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -42,7 +43,6 @@
     "eslint-plugin-mocha": "^6.2.2",
     "eslint-plugin-prettier": "^3.1.2",
     "grpc-tools": "1.8.1",
-    "grpc_tools_node_protoc_ts": "2.5.10",
```
hbergren: Why did this dependency get dropped?

Stormtv: @hbergren This dependency generated the typings for gRPC-Node output where as gRPC-Web does this automatically when using the `commonjs+dts` option in the generator.

PR #182, opened 2020-02-12
Update ripple-keypairs

0xCLARITY:

`ripple-keypairs` hit `1.0.0`, so this library should use that over the pre-1.0 version.

### Context of Change

The reason I wrote a PR instead of just using the dependabot PR is so I could rip out the `types/ripple-keypairs/index.d.ts` file, and so we could pull in the official typings now shipped with `ripple-keypairs`.

I also verified there were no TypeScript errors as a result of having official typings. We were already compatible, which is a good sign.

### Type of Change

## Before / After

Literally removed `.d.ts` file and updated the version in `package.json`.

## Test Plan

Ran tests locally, I assume CI will also pass.


Stormtv (approved):
🔥

keefertaylor (approved):

PR #187, opened 2020-02-19
Support for xrpl-secret-numbers encoded seed

WietseWind:
Add wallet with '[secret numbers](https://github.com/WietseWind/xrpl-secret-numbers)' seed.

Add `generateWalletFromSecretNumbers` and tests relying on [xrpl-secret-numbers](https://www.npmjs.com/package/xrpl-secret-numbers). Based on the idea by @nbougalis

### Context of Change

As it is really easy for humans to make mistakes in alpha chars (mnemonic, but family seeds are even worse) a new suggested seed encoding ("secret numbers") with per chunk checksumming will be used in [XUMM](https://xumm.app) by default. This commit adds support for 'secret numbers'. No new sub-dependencies as it relies on existing libs like ripple-keypairs / brorand. 

(Secret numbers are 'backwards compatible' with family seeds as they simply generate the seedbuffer usable encoded into the family seed)

(Type: `string` / `string[]` » `string[]` instead of `number[]` as a chunk of numbers can start with a zero)

### Type of Change

Add generateWalletFromSecretNumbers() & tests, add support for another secret type (mnemonic, family seed, adding 'secret numbers')

## Test Plan

- [X] Added tests
- [X] Passed existing tests
- [X] Linting OK,
Etc.


Stormtv (approved):
Nice, looks really good! Read over the source code of secret numbers and everything looks good. Made one issue over there but nothing major.

I do hope we can standardize on one format for exporting and displaying seeds. Also it would be nice if we could generate seeds using only one lib (even though they both use brorand at the lowest level.)

keefertaylor:
I'm generally happy with this and the implementation looks excellent (I'd expect no less). 

I just want to make sure that Xpring is ready to support this across the suite of products we're going to offer (ripple-lib and rippled, for starters). Can you clarify a bit about what you hope to do with this? Where else does this exist in the ripple ecosystem? No wrong answers, just fact finding.

WietseWind:
@Stormtv Thanks. Fixed that issue and updated this PR. 

@keefertaylor Great :) Regarding rippled: it's bad practice to send the family seed (or any other secret) anyway, and for now the secret numbers are compatible with a family seed anyway. 

@Stormtv + @keefertaylor regarding standardize & ripple-lib: I think at some point -if we decide secret numbers are helpful- I may suggest a PR for ripple-keypairs to add it and then another PR to ripple-lib, etc. 

One could argue that a semi XRPL-sepecific secret, be it family seed or secret numbers, is undesirable anyway as the BIP standards are more commonly used for hardware & software wallets & can be used to derive accounts for multiple currencies. 

However: I've had plenty mnemonic recovery requests over time. In all cases I was able to brute force recover the right mnemonic. Always spelling mistakes / handwriting errors / vowels simply replaced ending up with another existing, valid word. I do believe using the secret numbers representation will save a lot of people the fear of not being able to access their funds in the future. So `clarify a bit about what you hope to do with this`: free "our" ecosystem from typo/spelling mistakes/word confusion/corrupted family seeds. Everybody is familiar with number number ranges. A lot less intimidating for non-crypto users too.

intelliot (approved):

keefertaylor:
> Great :) Regarding rippled: it's bad practice to send the family seed (or any other secret) anyway, and for now the secret numbers are compatible with a family seed anyway.

Sure. I think that I'm concerned that the native methods in rippled (ex [wallet_propose](https://xrpl.org/wallet_propose.html)) return b58check encoded seeds rather than magic numbers. Similarly, the XRP {Dev, Test}Net faucets only deal in b58check encoded seeds and I'm sure theres some other tooling elsewhere as well. 

> regarding standardize & ripple-lib: I think at some point -if we decide secret numbers are helpful- I may suggest a PR for ripple-keypairs to add it and then another PR to ripple-lib, etc.

My concern here is that if we're not standardized across the entire ecosystem then we increase the number of options folks have and also the chances of confusion / loss of funds. 

> However: I've had plenty mnemonic recovery requests over time. In all cases I was able to brute force recover the right mnemonic. Always spelling mistakes / handwriting errors / vowels simply replaced ending up with another existing, valid word.

Argument bought

>  I sure hope Xpring & libs will attract more & new devs, opposed to a lot of existing devs using the other libs. I figured this is the place to allow devs new to the XRPL to get acquainted with the secret numbers (hopefully one day) standard.

-------

In general, I'd like to see this drafted as a standard in the XRP ecosystem before we fully support it. It sounds like there are numerous advantages, and I just want to make sure we have broad community support before we bolt this onto the side of the ecosystem and you / me / whoever else has to support it forever. 

Would you be willing to put out a draft standard?

WietseWind:
@keefertaylor Took a while (sorry, ended up at my "pile of todo's not written down").

- Draft XLS-12d: https://github.com/xrp-community/standards-drafts/issues/15
- Updated this PR with the most recent version of `master`.

amiecorso (approved):

PR #193, opened 2020-02-28
Upgrade rippled protos

keefertaylor:

Upgrade xpring-common-js to consume the latest protocol buffers in rippled, which contain breaking changes introduced in https://github.com/ripple/rippled/pull/3254. 

### Context of Change

No logic changes are introduced. 

Changes in the protocol buffers are:
- Change folder / package location of files
- Use numeric strings instead of numbers to prevent overflows when working with 64 bit integers. 
- Introduce new layers of abstraction internally. 

### Type of Change

## Before / After

Clients who use new versions of protocol buffers can now use the latest versions. 

## Test Plan
CI

### DIFF
--- a/src/serializer.ts
+++ b/src/serializer.ts
@@ -1,8 +1,11 @@
 import { Payment as LegacyPayment } from './generated/legacy/payment_pb'
 import { Transaction as LegacyTransaction } from './generated/legacy/transaction_pb'
 import { XRPAmount } from './generated/legacy/xrp_amount_pb'
-import { XRPDropsAmount } from './generated/rpc/v1/amount_pb'
-import { Payment, Transaction } from './generated/rpc/v1/transaction_pb'
+import { XRPDropsAmount } from './generated/org/xrpl/rpc/v1/amount_pb'
+import {
+  Payment,
+  Transaction,
+} from './generated/org/xrpl/rpc/v1/transaction_pb'
 import Utils from './utils'
 
 interface PaymentJSON {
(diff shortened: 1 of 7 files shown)


src/serializer.ts
```diff
@@ -44,16 +47,29 @@ class Serializer {
...
+      object.LastLedgerSequence = lastLedgerSequence
+    }
 
     // Convert account field, handling X-Addresses if needed.
     const accountAddress = transaction.getAccount()
     if (!accountAddress) {
       return
     }
-    const account = accountAddress.getAddress()
+    const account = accountAddress.getValue()?.getAddress()
```
hbergren: What's with the `.getValue()` everywhere? Is that part of the new Protobuf?

src/serializer.ts
```diff
@@ -44,16 +47,29 @@ class Serializer {
...
+      .getSigningPublicKey()
+      ?.getValue_asU8()
+    if (signingPubKeyBytes) {
+      object.SigningPubKey = Utils.toHex(signingPubKeyBytes)
+    }
+
+    const lastLedgerSequence = transaction.getLastLedgerSequence()?.getValue()
+    if (lastLedgerSequence) {
+      object.LastLedgerSequence = lastLedgerSequence
+    }
```
hbergren: Before the object always had a `LastLedgerSequence` property, and now it's conditional. Why is that?

test/serializer-test.ts
```diff
@@ -41,12 +50,12 @@ const accountXAddress = 'X7vjQVCddnQ7GCESYnYR3EdpzbcoAMbPw7s2xv8YQs94tv4'
...
-  fee: number,
-  lastLedgerSequence: number,
-  sequence: number,
-  account: string | undefined,
+  value: string,
+  destinationAddress: string,
+  fee: string,
+  lastLedgerSequenceNumber: number,
+  sequenceNumber: number,
+  senderAddress: string | undefined,
```
hbergren: Nice job using this breaking change opportunity to use some better names!

test/serializer-test.ts
```diff
@@ -55,28 +64,47 @@ function makeTransaction(
...
+  const destinationAccountAddress = new AccountAddress()
+  destinationAccountAddress.setAddress(destinationAddress)
+
+  const destination = new Destination()
+  destination.setValue(destinationAccountAddress)
 
   const payment = new Payment()
-  payment.setDestination(destinationAddress)
-  payment.setAmount(currencyAmount)
+  payment.setDestination(destination)
```
hbergren: This now takes a `Destination` instead of a `AccountAddress`? What is a `Destination`?

0xCLARITY (commented):
Seems reasonable.

I have a few questions below.

src/serializer.ts
keefertaylor: The old code was probably incorrect because the default value of protocol buffers numbers are 0.  We shouldn't require folks to set this optional property, so if it isn't set we omit if from the  JSON object. 

I'm applying this same logic to the above sequence and signing public key properties.

src/serializer.ts
keefertaylor: Yeah, there's a new `value` wrapper on a bunch of common types that added a layer of abstraction which was useful in the network handling layer.

test/serializer-test.ts
keefertaylor: A destination is just a wrapper for an account address which is specifically for a destination address. 

It's just another layer of abstraction introduced to make it easy to work with the rpc code on the node.

0xCLARITY (approved):
LGTM. Thanks for explaining all the changes!

Stormtv (approved):
LGTM

PR #204, opened 2020-03-07
[4.2.0] Sign raw JSON transactions and cut release

ledhed2222:

Added a new public function: `Signer.signTransactionFromJSON`. This allows non-JS runtimes to sign transactions without making network requests from JS or passing complex proto objects to JS. Instead, we can just pass the basic necessary JSON for signing.

Refactors `Signer.signTransaction` in terms of the above function.

(I believe) corrects the `TransactionJSON` type in `Serializer` so that it more accurately reflects the true JSON serialization of a transaction.

Adds a new test for the new `Signer` function.

NOTE: previously `Signer.signTransaction` would serialize the transaction object, then sign it, then serialize it again. **As far as I can tell** this is unnecessary so I removed it. It will now serialize the transaction object, sign it, then simply add the signature key to the already-serialized object. Would appreciate someone double-checking that!

### Context of Change

This will allow a (soon-to-be-released) Ruby Xpring library to work more easily.

### Type of Change

## Before / After

I think the above summary describes this.

## Test Plan

The automated tests still work. Added a new automated test. Verified that the generated code will do what is required for the upcoming Ruby lib.

### DIFF
--- a/src/serializer.ts
+++ b/src/serializer.ts
@@ -15,7 +15,7 @@ interface PaymentJSON {
   TransactionType: string
 }
 
-interface TransactionJSON {
+interface BaseTransactionJSON {
   Account: string
   Fee: string
   LastLedgerSequence: number
@@ -23,6 +23,15 @@ interface TransactionJSON {
   SigningPubKey: string
   TxnSignature?: string
 }
+
+interface PaymentTransactionJSONAddition extends PaymentJSON {
+  TransactionType: 'Payment'
+}
+
+type PaymentTransactionJSON = BaseTransactionJSON &
+  PaymentTransactionJSONAddition
+
+export type TransactionJSON = BaseTransactionJSON | PaymentTransactionJSON
 
 /**
  * Provides functionality to serialize from protocol buffers to JSON objects.
--- a/test/signer-test.ts
+++ b/test/signer-test.ts
@@ -150,24 +150,14 @@ describe('signer', function(): void {
       'X7vjQVCddnQ7GCESYnYR3EdpzbcoAMbPw7s2xv8YQs94tv4',
     )
 
-    // make mocha and typescript happy
-    if (destinationAddress == null) {
-      assert.fail('destinationAddress is undefined')
-      return
-    }
-    if (sourceAddress == null) {
-      assert.fail('sourceAddress is undefined')
-      return
-    }
-
     const transactionJSON = {
-      Account: sourceAddress.address,
+      Account: sourceAddress!.address,
       Fee: '10',
       Sequence: 1,
       LastLedgerSequence: 0,
       SigningPubKey: 'BEEFDEAD',
       Amount: '1000',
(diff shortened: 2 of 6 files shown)


src/serializer.ts
```diff
@@ -24,6 +24,15 @@ interface TransactionJSON {
   TxnSignature?: string
 }
 
+interface PaymentTransactionJSONAddition extends PaymentJSON {
+  TransactionType: 'Payment'
```
keefertaylor: I think you should include the amount and destination field here?

test/signer-test.ts
```diff
@@ -138,4 +138,60 @@ describe('signer', function(): void {
...
+    const wallet = new FakeWallet(fakeSignature)
+    const destinationAddress = Utils.decodeXAddress(
+      'XVPcpSm47b1CZkf5AkKM9a84dQHe3m4sBhsrA4XtnBECTAc',
+    )
+    const sourceAddress = Utils.decodeXAddress(
+      'X7vjQVCddnQ7GCESYnYR3EdpzbcoAMbPw7s2xv8YQs94tv4',
+    )
+
+    // make mocha and typescript happy
+    if (destinationAddress == null) {
```
keefertaylor: You can just append a `!` on line 148 / 151 to force unwrap these objects.

keefertaylor (commented):
Sorry for latency on this review. One question about the type formatting for JSON and I'm happy.

keefertaylor:
Could you also update `CHANGELOG.md` with a record of this new method?

keefertaylor:
\+ @hbergren as a secondary reviewer.

src/serializer.ts
```diff
@@ -24,6 +24,15 @@ interface TransactionJSON {
...
+}
+
+type PaymentTransactionJSON = BaseTransactionJSON &
+  PaymentTransactionJSONAddition
+
+export type TransactionJSON = BaseTransactionJSON | PaymentTransactionJSON
```
hbergren: If the `amount` and `destination` fields should be included, I think this makes sense.

If they shouldn't, maybe just add an optional `TransactionType` to the `TransactionJSON` object?

0xCLARITY (approved):
LGTM! Nice and small diff.

src/serializer.ts
ledhed2222: it is included, by virtue of extending `paymentJSON` :)

src/serializer.ts
ledhed2222: i think they should be included, to show that to have a payment-flavored transaction you need those additional fields

Stormtv (approved):

keefertaylor (approved):

keefertaylor:
This includes a release cut. I'm going to run the following commands:

After merging this PR I will run:

```
$ git checkout master && git pull
$ git tag '4.2.0' && git push --tags
$ npm publish
```

PR #236, opened 2020-04-10
Modify PayID Format

keefertaylor:

Migrate PayID format from:
`$xpring.money/georgewashington`
to
`georgewashington$xpring.money`

### Context of Change

Product requested change in format. 

Remove references to 'payment pointer' and rename a few classes / methods to reflect that we're not longer compliant with the payment pointer spec.

This is a breaking change, but PayID is experimental so I don't consider it a big deal.

### Type of Change

- [ ] Documentation Updates
- [ ] Release

## Before / After

New spec implemented.

## Test Plan

New tests provided for CI. 

## Future Tasks
Uptake parsing logic and test cases in client libraries (JS / Swift / Java)


src/PayID/pay-id-components.ts
```diff
@@ -0,0 +1,15 @@
+/**
+ * A class which which encapsulates components of a Pay ID.
```
amiecorso, line 2: s/which which/which

test/PayID/pay-id-utils-test.ts
```diff
@@ -1,66 +1,66 @@
...
 
 describe('PayIDUtils', function(): void {
-  it('parse payment pointer - host and path', function(): void {
-    // GIVEN a payment pointer with a host and a path.
-    const rawPaymentPointer = '$example.com/foo'
+  it('parse Pay ID - valid', function(): void {
+    // GIVEN a Pay ID with a host and a path.
+    const host = 'xpring.money'
+    const path = 'georgewashington'
+    const rawPayID = `${path}$${host}`
```
amiecorso, line 10: is it just me or does host and path seem flipped here?

amiecorso: oh nevermind, I see the whole point of this PR

src/PayID/pay-id-components.ts
```diff
@@ -0,0 +1,15 @@
...
+ */
+export default class PayIDComponents {
+  /**
+   * Create a new PayIDComponents.
+   *
+   * @param host The host of the payment pointer, starting with a leading '/'.
```
hbergren, line 8: Does host start with a `leading '/'`, or does `path`?

test/PayID/pay-id-utils-test.ts
```diff
@@ -1,66 +1,66 @@
...
-    assert.equal(paymentPointer?.path, '/foo')
+    assert.equal(payIDComponents?.host, host)
+    assert.equal(payIDComponents?.path, `/${path}`)
   })
 
-  it('parse payment pointer - well known path', function(): void {
-    // GIVEN a payment pointer with a well known path.
-    const rawPaymentPointer = '$example.com'
+  it('parse Pay ID - multiple dollar signs', function(): void {
+    // GIVEN a Pay ID with too many '$'.
```
hbergren, line 21: I'm not sure this test is accurate. You _can_ have `$`s in the `path` of a URL. Maybe we should say the PayID spec disallows that, but I'm not sure it does right now.

Thoughts @austin-king ?

src/PayID/pay-id-components.ts
keefertaylor: Done

austin-king:
Can we hold off on implementation changes until we know how we're changing the format? Appreciate the speed but I don't want to waste your time if you have to reimplement some things.

For example, in this you parse user and host from this and treat the user as a path, we're not sure if we're going to stick with the payment pointer URL format behind the scenes yet

test/PayID/pay-id-utils-test.ts
dino-rodriguez: I think `$` should be a reserved char

austin-king: Think it's best to disallow multiple $'s for PayID. I imagine this would be very confusing to any user.

austin-king:
> @tedkalaw this PR looks good except we are only slicing off the last `/` to get the user.
> AKA for the URL https://wallet.com/payid/alice this PR would return:
> payid/alice$wallet.com
> but what we need is:
> alice$wallet.com/payid
> 
> Once we get that fixed we can merge this
> Idk what is best practice if Keefer is out today, should we just push to his branch?

@tedkalaw I read this PR wrong, the above comment is inaccurate. It provides the functionality as desired b/c it only concerns itself with mapping PayID -> URL not URL -> PayID. 

Approving

tedkalaw:
i'm commandeering this ship

PR #264, opened 2020-05-06
Xpring ESLint (1/X)

hbergren:

I made a few changes to bring us closer to alignment with the linting rules of `@xpring-eng/eslint-config-base`:

- Import statements are sorted, and have spaces between different import "groups"
- No 1-liner if statements
- Add `readonly` to applicable class properties
- Use the safer `??` operator over `||` where possible
- Remove types from JSDoc comments because we're using TypeScript, so we already have that information
- Add `-` hyphens after `@param`s in JSDocs
- Use unicode regexes
- Convert `test/*-test.ts` files to `test/*.test.ts` files.

### Context of Change

I wrote a [`@xpring-eng` ESLint config](https://github.com/xpring-eng/eslint/tree/master/packages/eslint-config-base) and published it to NPM, and I'd like to migrate over our other TS projects to using it.

In general, I think it will help us write code that is more type-safe and higher quality.

### Type of Change

- [ ] Documentation Updates
- [ ] Release

## Test Plan

`npm run test` still works.


.mocharc.js
```diff
@@ -8,8 +8,8 @@ module.exports = {
   extension: [
     'ts'
   ],
-  spec: "test/**/*-test.ts",
-  
+  spec: "test/**/*.test.ts",
```
hbergren, line 11: In general, our other projects seem to favor `*.test.ts` files to distinguish test files over `*-test.ts`.

I actually baked this assumption into the `@xpring-eng/eslint-config-base`, so there are a bunch of test-file specific overrides using the `*.test.ts` pattern.

So, I just went ahead and migrated the test files to that naming convention.

src/Common/utils.ts
```diff
@@ -1,8 +1,9 @@
 import { createHash } from 'crypto'
+
 import * as addressCodec from 'ripple-address-codec'
```
hbergren, line 3: The Xpring ESLint config has rules about import ordering and spacing.

Because `crypto` is a module built-in to Node, and `ripple-address-codec` is an external dependency, they are considered to be different "import groups" and thus require a blank line between them.

src/Common/utils.ts
```diff
@@ -29,7 +30,7 @@ class Utils {
    * This function returns true for both X-addresses and classic addresses.
    * @see https://xrpaddress.info/
    *
-   * @param address An address to check.
+   * @param address - An address to check.
```
hbergren, line 33: This syntax (hyphen separating parameter name from description) for `@param` tags is enforced by the `tsdoc` ESLint plugin. I think this makes things a little more readable though.

src/Common/utils.ts
```diff
@@ -170,12 +171,12 @@ class Utils {
...
    * Check if the given string is valid hex.
    *
-   * @param input The input to check.
+   * @param input - The input to check.
    * @returns true if the input is valid hex, otherwise false.
    */
   public static isHex(input: string): boolean {
-    const hexRegEx = /([0-9]|[a-f])/gim
-    return (input.match(hexRegEx) || []).length === input.length
+    const hexRegEx = /([0-9]|[a-f])/gimu
```
hbergren, line 178: So, apparently, JS regexes have a bunch of wonky behavior unless you enable the `u` unicode flag:

https://eslint.org/docs/rules/require-unicode-regexp

src/Common/utils.ts
```diff
@@ -170,12 +171,12 @@ class Utils {
...
+    return (input.match(hexRegEx) ?? []).length === input.length
```
hbergren, line 179: Use `??` nullish operator where appropriate.

src/PayID/pay-id-utils.ts
```diff
@@ -7,20 +7,26 @@ export default class PayIDUtils {
...
    *
-   * @param parsePayID The input Pay ID.
+   * @param parsePayID - The input Pay ID.
    * @returns A PayIDComponents object if the input was valid, otherwise undefined.
    */
   public static parsePayID(payID: string): PayIDComponents | undefined {
-    if (!PayIDUtils.isASCII(payID)) return
+    if (!PayIDUtils.isASCII(payID)) {
+      return
+    }
```
hbergren, line 16: I wrote a rule banning single line `if` statements.

I go back and forth, but in general, I think it's reasonable to be consistent, even if this is a little more verbose in terms of LoC.

src/XRP/serializer.ts
```diff
@@ -37,9 +38,9 @@ class Serializer {
   /**
    * Convert a Transaction to a JSON representation.
    *
-   * @param {proto.Transaction} transaction A Transaction to convert.
-   * @param signature An optional hex encoded signature to include in the transaction.
-   * @returns {Object} The Transaction as JSON.
+   * @param transaction - A Transaction to convert.
```
hbergren, line 41: The TSDoc ESLint plugin bans the use of types in the JSDoc comments, because we're using TypeScript and we already have that information available to us.

test/XRP/fakes/fake-wallet.ts
```diff
@@ -19,10 +19,10 @@ class FakeWallet extends Wallet {
   /**
    * Initialize a wallet which will always produce the same signature when asked to sign a string.
    *
-   * @param {String} signature The signature this wallet will produce.
+   * @param signature - The signature this wallet will produce.
    */
   constructor(
-    private signature: string,
+    private readonly signature: string,
```
hbergren, line 25: Because this is a `private` class member that is never mutated, some rule we have is smart enough to realize that this can be marked `readonly`.

It couldn't do this for a `public` or `protected` member obviously.

src/XRP/signer.ts
```diff
@@ -11,9 +13,9 @@ class Signer {
   /**
    * Encode the given raw JSON transaction to hex and sign it.
    *
-   * @param {TransactionJSON} transactionJSON The raw transaction JSON object.
-   * @param {Wallet} wallet The wallet to sign the transaction with.
-   * @returns {Uint8Array} A set of bytes representing the inputs and a signature.
+   * @param transactionJSON - The raw transaction JSON object.
+   * @param wallet  -The wallet to sign the transaction with.
```
amiecorso, line 17: should there be a space both before and after the hyphen?

hbergren: Good catch! That's a typo on my end. This rule should really be auto-fixable, but the `tsdoc` plugin is in its early days, so I had to hand-fix these.

PR #265, opened 2020-05-06
Try out Github Actions for cutting releases

hbergren:

I threw some comments into the code below to explain what's going on.

But, basically this adds two new Github Actions to this repo:

- Cut a Github Release when you push up a `git tag`
- Publish to NPM whenever a new Github Release is cut

### Context of Change

We want to make deploying the SDK easier, and this seems like a reasonable way to go about doing that.

### Type of Change

- [ ] Documentation Updates
- [ ] Release

## Before / After

Stuff might happen automagically when you push a local version tag?

## Test Plan

???

I have no idea how to actually test Github Actions. (To be fair, I put zero effort into finding out).

Maybe we should just merge this and try creating a `tag` locally and pushing it up the next time we want to cut a release and see what happens?


.github/workflows/github-release.yml
```diff
@@ -0,0 +1,27 @@
+# https://github.com/marketplace/actions/create-a-release
+on:
+  push:
+    # Sequence of patterns matched against refs/tags
+    tags:
+      - 'v*' # Push events to matching v*, i.e. v1.0, v20.15.10
```
hbergren, line 6: I totally stole this from the base Github example, but _basically_, I think the way this works is you:

- Cut a `git tag` locally
- Push up your `git tag` to Github
- Github will cut a release with the name of the tag that you cut locally

.github/workflows/github-release.yml
```diff
@@ -0,0 +1,27 @@
...
+        uses: actions/checkout@v2
+      - name: Create Release
+        id: create_release
+        uses: actions/create-release@v1
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }} # This token is provided by Actions, you do not need to create your own token
+        with:
+          tag_name: ${{ github.ref }}
+          release_name: ${{ github.ref }}
+          body: ${{ github.event.head_commit.message }}
```
hbergren, line 25: I _think_ this syntax is correct. I'm trying to give the release body the body of the most recent commit.

.github/workflows/npm-publish.yml
```diff
@@ -0,0 +1,45 @@
+# This workflow will run tests using node and then publish a package to GitHub Packages when a release is created
+# For more information see: https://help.github.com/actions/language-and-framework-guides/publishing-nodejs-packages
+
+name: Node.js Package
+
+on:
+  release:
+    types: [created]
```
hbergren, line 8: I think the `github-release` Action should trigger this one, because that creates a Github release.

.github/workflows/npm-publish.yml
```diff
@@ -0,0 +1,45 @@
...
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-node@v1
+        with:
+          node-version: 10
+      - run: npm run test
```
hbergren, line 18: Seems reasonable to test our code before publishing to NPM.

.github/workflows/npm-publish.yml
```diff
@@ -0,0 +1,45 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-node@v1
+        with:
+          node-version: 10
+          registry-url: https://registry.npmjs.org/
+      - run: npm publish
+        env:
+          NODE_AUTH_TOKEN: ${{secrets.NPM_TOKEN}}
```
hbergren, line 31: Only thing for you to do Keefer is to add a `NPM_TOKEN` here: https://github.com/xpring-eng/xpring-common-js/settings/secrets

Then this should be able to publish `@xpring-eng` packages.

.github/workflows/npm-publish.yml
```diff
@@ -0,0 +1,45 @@
...
+
+  publish-gpr:
```
hbergren, line 33: I have no idea if we need/want this.

This publishes the package to the Github package registry. We can just remove all these lines if we don't care.

keefertaylor:
This looks really solid. 

A few things:
- The config for `npm publish` already runs `npm test` so I think the test phase is redundant. 
- I usually tag releases as `1.0.0` instead of `v1.0.0` but I don't care too much. Is `v*` standard? 
- I don't care about GitHub package repo but I don't mind if we do publish there. Do npm developers look there? Should we be doing this? 

Adding @tlongwell-ripple because I think he'll have comments about:
- Storing a secret on GitHub and if we think that's reasonably secure (my answer is looks good to me) 
- Whether he is just going to migrate all of this to GitLab and make this code redundant
- Any other dev-ops-y things I haven't thought about.

hbergren:
> The config for npm publish already runs npm test so I think the test phase is redundant.

Reasonable, I'll cut that then.

> I usually tag releases as 1.0.0 instead of v1.0.0 but I don't care too much. Is v* standard?

Apparently. I did not know this, but apparently the `vX.Y.Z` tag syntax [is defined in the SemVer spec](https://semver.org/spec/v1.0.0.html#tagging-specification-semvertag).

> I don't care about GitHub package repo but I don't mind if we do publish there. Do npm developers look there? Should we be doing this?

No idea. It's weird that Github owns both NPM and the Github Package Registry. I don't see the harm in publishing to GPR, but I don't see any real benefit either.

hbergren:
Bump @keefertaylor , @tlongwell-ripple

hbergren:
I'm going to close this so it's not perpetually in my open PRs list, but I think we should explore this at some point in the future.

PR #267, opened 2020-05-11
Introduce Wallet Factory

keefertaylor:

Create a `WalletFactory` Class. 

### Context of Change

There are a number of ways to generate wallets. Currently, these exist as static methods on the wallet class, since JavaScript only lets a class have one constructor. 

This leads to two problems:
- In JS, developers sometimes can't find the method they need because they are looking at constructors
- In Swift / Java classes, functionality is mismatched and hard to follow as instance methods of a `Wallet` object use a static methods of an underlying JavaScript class.

Implementing a `WalletFactory` will aggregate this functionality as we add additional wallet constructor methods.

This PR wires an initial factory method and then returns 

### Type of Change

- [ ] Documentation Updates
- [ ] Release

## Before / After

A new interface provides a logical way to generate wallets, rather than stuffing generation logic onto the class itself. 

## Test Plan

New tests provided for CI. 

## Future Tasks
- Migrate additional methods
- Deprecate existing methods (And remove, some time in the future)
- Consider making Wallet's initializer private


src/index.ts
```diff
@@ -1,16 +1,15 @@
-export { Transaction } from './XRP/generated/org/xrpl/rpc/v1/transaction_pb'
-
-export { default as Wallet } from './XRP/wallet'
+// Exported Common Functionality.
```
keefertaylor, line 1: Note I re-ordered this file by:
- Component (alphabetically: Common, PayID, XRP)
- Alphabetically line by line

amiecorso (approved):

keefertaylor:
Ping @stephengu

stephengu (approved):
wow.. shame on me for missing this review... my bad.
so i'm assuming some of the new constructors to be added would be like wallet from seed, etc?

keefertaylor:
> so i'm assuming some of the new constructors to be added would be like wallet from seed, etc?
Correct. Was going to migrate over slowly for ease of review.

keefertaylor:
ping @LoisRP

LoisRP (approved):
Approved.

PR #268, opened 2020-05-11
Provide Idiomatically Cased Pay ID Classes

keefertaylor:

JavaScript tends to favor only casing the first letter of acronyms (PayId vs PayID). Start to re-case our classes to this style. 

### Context of Change

We use both styles of this casing across our JavaScript libraries (see `IlpClient` and `XRPClient`). JavaScript tends to favor the former case. 

This PR re-cases the Pay ID functionality to use the correct casing. Specifically:
- Duplicate each incorrectly cased class into a correctly cased class. 
- Modify the incorrectly cased class to call the correctly case classed
- Mark the incorrectly cased class as deprecated
- Update our tests immediately to the new casing
- Update `index.ts` to export both classes. 

A future PR will remove the incorrectly cased classes in N+2 releases, as described by our new release policy. 

### Type of Change

- [ ] Documentation Updates
- [ ] Release

## Before / After

A bridge to correctly named classes is provided. 

## Test Plan

CI

### DIFF
--- a/src/PayID/pay-id-components.ts
+++ b/src/PayID/pay-id-components.ts
@@ -1,7 +1,29 @@
+// Disable multiple classes to accommodate the switch to idiomatic style naming.
+// TODO(keefertaylor): Remove this when migration is complete.
+/* eslint-disable  max-classes-per-file */
+
 /**
  * A class which encapsulates components of a Pay ID.
  */
-export default class PayIDComponents {
+export default class PayIdComponents {
+  /**
+   * Create a new PayIdComponents.
+   *
+   * @param host - The host of the payment pointer.
+   * @param path - The path of the payment pointer, starting with a leading '/'.
+   */
+  public constructor(
+    public readonly host: string,
+    public readonly path: string,
+  ) {}
+}
+
+/**
+ * A class which encapsulates components of a Pay ID.
+ *
+ * @deprecated Please use the idiomatically named `PayIdComponents` class instead.
+ */
+export class PayIDComponents {
   /**
    * Create a new PayIDComponents.
    *
--- a/src/PayID/pay-id-utils.ts
+++ b/src/PayID/pay-id-utils.ts
@@ -1,17 +1,21 @@
-import PayIDComponents from './pay-id-components'
+// Disable multiple classes to accommodate the switch to idiomatic style naming.
+// TODO(keefertaylor): Remove this when migration is complete.
+/* eslint-disable  max-classes-per-file */
+
+import PayIdComponents, { PayIDComponents } from './pay-id-components'
 
 /**
- * A static utility class for PayID.
+ * A static utility class for Pay ID functionality.
  */
-export default class PayIDUtils {
+export default class PayIdUtils {
   /**
    * Parse a PayID string to a set of PayIDComponents object
    *
    * @param parsePayID - The input Pay ID.
    * @returns A PayIDComponents object if the input was valid, otherwise undefined.
    */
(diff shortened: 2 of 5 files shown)


src/PayID/pay-id-components.ts
```diff
@@ -1,7 +1,29 @@
+// Disable multiple classes to accommodate the switch to idiomatic style naming.
+// TODO(keefertaylor): Remove this when migration is complete.
+/* eslint-disable  max-classes-per-file */
```
hbergren, line 3: Super helpful comment for this `eslint-disable`.

src/PayID/pay-id-utils.ts
```diff
@@ -49,3 +53,24 @@ export default class PayIDUtils {
...
+   * @param parsePayID - The input Pay ID.
+   * @returns A PayIDComponents object if the input was valid, otherwise undefined.
+   */
+  public static parsePayID(payID: string): PayIDComponents | undefined {
+    return PayIdUtils.parsePayID(payID)
+  }
+
+  /** Please do not instantiate this static utility class. */
+  // eslint-disable-next-line @typescript-eslint/no-empty-function
+  private constructor() {}
```
hbergren, line 75: Good comment.

Total aside: There's a `@typescript-eslint` rule that disallows using "classes as namespaces", which a class with only static methods is. I know that's a pretty common pattern in `xpring-common-js`, and I just wanted to get a feel for how you felt about that.

[Here is the actual rule documentation for some reasoning and examples](https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/no-extraneous-class.md).

src/PayID/pay-id-utils.ts
```diff
@@ -1,17 +1,21 @@
...
+import PayIdComponents, { PayIDComponents } from './pay-id-components'
 
 /**
- * A static utility class for PayID.
+ * A static utility class for Pay ID functionality.
  */
-export default class PayIDUtils {
+export default class PayIdUtils {
   /**
    * Parse a PayID string to a set of PayIDComponents object
```
amiecorso, line 12: should this line be `PayIdComponents` ?

src/PayID/pay-id-utils.ts
```diff
@@ -1,17 +1,21 @@
...
    *
    * @param parsePayID - The input Pay ID.
    * @returns A PayIDComponents object if the input was valid, otherwise undefined.
```
amiecorso, line 15: here too

test/PayID/pay-id-utils.test.ts
```diff
@@ -11,7 +11,7 @@ describe('PayIDUtils', function (): void {
     const rawPayID = `${path}$${host}`
 
     // WHEN it is parsed to components.
-    const payIDComponents = PayIDUtils.parsePayID(rawPayID)
+    const payIDComponents = PayIdUtils.parsePayID(rawPayID)
```
amiecorso, line 14: does it matter that this `payIDComponents` variable is cased in the old style?  I know the style applies mostly to classnames but wondering if we would benefit from that consistency wherever possible?

amiecorso: (here and throughout the file)

hbergren:
Random question, are `XRP` class and variable names still going to be in full cap casing, or will they be `Xrp` eventually?

keefertaylor:
> Random question, are XRP class and variable names still going to be in full cap casing, or will they be Xrp eventually?

We talked offline. Aiming for Xrp since we can enforce with tooling.

PR #272, opened 2020-05-12
Xpring eslint 2

hbergren:

Various linting changes to bring this repo closer to being able to use the `xpring-eslint` configuration.

In particular, this PR:
- Breaks up comments into multiple lines if they are too long
- Adds the `abstract` keyword to classes used as namespaces (static only classes)
- Sorts `public` class members before `private` class members
- Avoids using magic numbers like `32`
- Removes unused vars in `catch` blocks
- Enforces a naming convention for boolean variables
- Removes inline comments
- Enforces explicit accessibility modifiers on class members
- Enforces that comments need to be full sentences
- Updates outdated JSDocs

### Context of Change

Part of converting our repos to use the stricter `xpring-eslint`.

### Type of Change

- [ ] Documentation Updates
- [ ] Release

## Test Plan

`npm run test` still passes, because no functionality changed.

### DIFF
--- a/src/Common/utils.ts
+++ b/src/Common/utils.ts
@@ -177,7 +177,7 @@ abstract class Utils {
     const hashHex = sha512.update(bytes).digest('hex').toUpperCase()
     const hash = this.toBytes(hashHex)
 
-    return hash.slice(0, 32)
+    return hash.slice(0, hash.length / 2)
   }
 }
 
--- a/src/Common/utils.ts
+++ b/src/Common/utils.ts
@@ -10,7 +10,8 @@ import * as addressCodec from 'ripple-address-codec'
 const signedTransactionPrefixHex = '54584E00'
 
 /**
- * A simple property bag which contains components of a classic address. Components contained in this object are neither sanitized or validated.
+ * A simple property bag which contains components of a classic address.
+ * Components contained in this object are neither sanitized or validated.
  */
 export interface ClassicAddress {
   /** A classic address. */
--- a/src/Common/utils.ts
+++ b/src/Common/utils.ts
@@ -176,7 +176,7 @@ class Utils {
    * @returns true if the input is valid hex, otherwise false.
    */
   public static isHex(input: string): boolean {
-    const hexRegEx = /([0-9]|[a-f])/gimu
+    const hexRegEx = /(?:[0-9]|[a-f])/gimu
     return (input.match(hexRegEx) ?? []).length === input.length
   }
 }
--- a/src/Common/utils.ts
+++ b/src/Common/utils.ts
@@ -24,7 +24,7 @@ export interface ClassicAddress {
   test: boolean
 }
 
-class Utils {
+abstract class Utils {
   /**
    * Validate that the given string is a valid address for the XRP Ledger.
    *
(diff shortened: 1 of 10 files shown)


test/XRP/serializer.test.ts
```diff
@@ -20,7 +20,7 @@ import {
 } from '../../src/XRP/generated/org/xrpl/rpc/v1/transaction_pb'
 import Serializer from '../../src/XRP/serializer'
 
-/** Constants for transactions */
+/** Constants for transactions. */
```
hbergren, line 23: Comments need to be full sentences.

src/Common/utils.ts
```diff
@@ -165,18 +177,7 @@ class Utils {
...
-  /**
-   * Check if the given string is valid hex.
-   *
-   * @param input - The input to check.
-   * @returns true if the input is valid hex, otherwise false.
-   */
-  public static isHex(input: string): boolean {
-    const hexRegEx = /([0-9]|[a-f])/gimu
-    return (input.match(hexRegEx) ?? []).length === input.length
+    return hash.slice(0, hash.length / 2)
```
hbergren, line 180: Keefer suggested `hash.length / 2` over 32 to avoid using magic numbers, and to better explain what's going on here.

test/XRP/serializer.test.ts
```diff
@@ -40,11 +40,11 @@ const accountXAddress = 'X7vjQVCddnQ7GCESYnYR3EdpzbcoAMbPw7s2xv8YQs94tv4'
...
  * @param value - The amount of XRP to send, in drops.
- * @param destination - The destination address.
+ * @param destinationAddress - The destination address.
  * @param fee - The amount of XRP to use as a fee, in drops.
- * @param lastLedgerSequence - The last ledger sequence the transaction will be valid in.
- * @param sequence - The sequence number for the sending account.
- * @param account - The address of the sending account.
+ * @param lastLedgerSequenceNumber - The last ledger sequence the transaction will be valid in.
+ * @param sequenceNumber - The sequence number for the sending account.
+ * @param senderAddress - The address of the sending account.
```
hbergren, line 47: The parameter names in the JSDoc here were out of date.

test/XRP/wallet.test.ts
```diff
@@ -5,7 +5,8 @@ import 'mocha'
 
 /**
  * A mapping of input and expected outputs for BIP39 and BIP44.
- * @see https://iancoleman.io/bip39/#english
+ *
+ * @see [BIP39 - Mnemonic Code Converter]{@link https://iancoleman.io/bip39/#english}
```
hbergren, line 9: The `@see` tag is kind of weird in JSDoc and it's nondeterministic what will happen. TSDoc recommends using `@see` with the `@link` tag to be clear about what is going on.

src/Common/utils.ts
```diff
@@ -154,6 +155,17 @@ class Utils {
...
+  /**
+   * Check if the given string is valid hex.
+   *
+   * @param input - The input to check.
+   * @returns true if the input is valid hex, otherwise false.
+   */
+  public static isHex(input: string): boolean {
+    const hexRegEx = /(?:[0-9]|[a-f])/gimu
+    return (input.match(hexRegEx) ?? []).length === input.length
+  }
```
hbergren, line 167: We have a rule that enforces `member-ordering` in classes, and [right now it's configured](https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/member-ordering.md#default-configuration) to put `public static` members before `private static` members.

I'm open to changing the ordering of class members if we have a different preference. Also, this rule allows enforcing an alphabetical sort order, which I currently have disabled, but we might want to enable.

test/XRP/fakes/fake-wallet.ts
```diff
@@ -21,7 +21,7 @@ class FakeWallet extends Wallet {
    *
    * @param signature - The signature this wallet will produce.
    */
-  constructor(
+  public constructor(
```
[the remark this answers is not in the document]
keefertaylor: Awesome. I prefer this, I thought the going style was to drop `public`. Happy to see this go in.

PR #282, opened 2020-05-18
Xpring ESLint - Round 3

hbergren:

- Avoid classes with only static members
- Avoid negative if conditions
- Disallow parameter properties for classes
- Consistently return stuff from functions that return
- Follow `member-ordering` for class members
- Avoid magic numbers

### Context of Change

3rd (and almost final) step of bringing `xpring-common-js` onto the new lint config.

### Type of Change

- [ ] Documentation Updates
- [ ] Release

## Test Plan

`npm run test` is all still green.


src/Common/utils.ts
```diff
@@ -24,7 +24,7 @@ export interface ClassicAddress {
   test: boolean
 }
 
-abstract class Utils {
+const utils = {
```
hbergren, line 27: Change from a class to an object that holds functions.

This is the proper fix for the [no-extraneous-class](https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/no-extraneous-class.md) rule. I did ask about [using abstract classes for this problem](https://github.com/typescript-eslint/typescript-eslint/issues/2015), but the consensus seems to be to just use EcmaScript modules.

It's a little odd that we're exporting an object as opposed to the individual functions, but this keeps us consistent with the old way of exporting the class.

src/Common/utils.ts
```diff
@@ -85,13 +85,13 @@ abstract class Utils {
     }
 
     // Xpring Common JS's API takes a string|undefined while the underlying address library wants string|false.
-    const shimTagParameter = tag !== undefined ? tag : false
+    const shimTagParameter = tag ?? false
```
hbergren, line 88: There's [an ESLint rule](https://eslint.org/docs/rules/no-negated-condition) against `if(negative) {...} else {...}`, because there's usually a better way to write it (like here).

src/PayID/pay-id-components.ts
```diff
@@ -6,16 +6,19 @@
...
    * @param host - The host of the payment pointer.
    * @param path - The path of the payment pointer, starting with a leading '/'.
    */
-  public constructor(
-    public readonly host: string,
-    public readonly path: string,
-  ) {}
+  public constructor(host: string, path: string) {
+    this.host = host
+    this.path = path
```
hbergren, line 20: There's [a rule](https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/no-parameter-properties.md) that disallows declaring class properties in the constructor. I'm ok with disabling this rule for now, but since _some_ properties might not be declarable in the constructor, it makes sense to be consistent.

src/PayID/pay-id-utils.ts
```diff
@@ -1,39 +1,37 @@
...
-   * @param parsePayID - The input Pay ID.
+   * @param payID - The input Pay ID.
    * @returns A PayIDComponents object if the input was valid, otherwise undefined.
    */
-  public static parsePayID(payID: string): PayIdComponents | undefined {
-    if (!PayIdUtils.isASCII(payID)) {
-      return
+  parsePayID(payID: string): PayIdComponents | undefined {
+    if (!this.isASCII(payID)) {
+      return undefined
```
hbergren, line 17: The [consistent-return](https://eslint.org/docs/rules/consistent-return) rule requires being explicit with your `return` statements if your function contains a `return` statement that returns something other than undefined (like this function with `return new PayIdComponents()`.

src/XRP/wallet.ts
```diff
@@ -27,6 +27,23 @@ export interface WalletGenerationResult {
...
+  private readonly test: boolean
+
+  /**
+   * Create a new Wallet object.
+   *
+   * @param publicKey - The given public key for the wallet.
+   * @param privateKey - The given private key for the wallet.
+   * @param test - Whether the address is for use on a test network, defaults to `false`.
+   */
+  public constructor(publicKey: string, privateKey: string, test = false) {
```
hbergren, line 41: Moved the constructor because the [member-ordering](https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/member-ordering.md) rule defines an order that class members should appear in (and constructors come before any methods).

test/Common/utils.test.ts
```diff
@@ -162,13 +163,14 @@ describe('utils', function (): void {
       classicAddress!.address,
       'rU6K7V3Po4snVhBBaU29sesqs2qTQJWDw1',
     )
-    assert.strictEqual(classicAddress!.tag, 12345)
+    assert.strictEqual(classicAddress!.tag, tag)
```
hbergren, line 166: Added a variable for this to avoid the `no-magic-numbers` rule. This seems reasonable because the `GIVEN` comment includes the tag

tsconfig.json
```diff
@@ -26,9 +26,7 @@
     "allowJs": true,
     "forceConsistentCasingInFileNames": true,
     "esModuleInterop": true,
-    "typeRoots": ["./src/types", "./node_modules/@types"]
+    "typeRoots": ["@types", "./node_modules/@types"]
   },
-  "include": [
-    "src/**/*"
-  ]
+  "include": ["src/**/*"]
```
hbergren, line 31: Moved the `src/types` directory to `@types`. I _think_ this is more conventional, but let me know if there's a strong reason to have custom declaration files in the `src` directory.

src/PayID/pay-id-components.ts
keefertaylor: I'm happy to use lint rules if that's what folks like doing.

keefertaylor: The only reason I could think of not liking this style is that you forget to assign a variable. However, this appears to not be the case, since the following errors:

```
public constructor(host: string, path: string) {
    this.host = host
}
```

Yells taht I didn't initialize path.

src/PayID/pay-id-utils.ts
keefertaylor: woo. I actually prefer this.

keefertaylor:
Verified this works in XpringKit and Xpring4J. 

I'm going to avoid merging anything so that you can get this PR landed cleanly and I'll rebase the upstream PRs for any issues. Are there further linter PRs to go or is everything enabled?

hbergren:
There is probably one more linter PR to go.

The big blocker here is that `tsdoc` does not support the `@see` tag yet. I'm working on that here:
https://github.com/microsoft/tsdoc/pull/236

The only other violations are things like this:
```
/Users/hbergren/Documents/xpring-eng/xpring-common-js/src/XRP/serializer.ts
   45:3   warning  Method 'transactionToJSON' has too many lines (83). Maximum allowed is 50       max-lines-per-function
   45:20  warning  Method 'transactionToJSON' has a complexity of 14. Maximum allowed is 8         complexity
   45:20  warning  Method 'transactionToJSON' has too many statements (39). Maximum allowed is 10  max-statements
  135:16  warning  Method 'paymentToJSON' has too many statements (22). Maximum allowed is 10      max-statements

/Users/hbergren/Documents/xpring-eng/xpring-common-js/test/XRP/serializer.test.ts
  51:1  warning  Function 'makeTransaction' has too many parameters (7). Maximum allowed is 4    max-params
  51:1  warning  Function 'makeTransaction' has too many statements (34). Maximum allowed is 20  max-statements

/Users/hbergren/Documents/xpring-eng/xpring-common-js/test/XRP/signer.test.ts
  24:24  warning  Maximum number of dependencies (8) exceeded                   import/max-dependencies
  27:14  warning  Function has too many statements (39). Maximum allowed is 20  max-statements

/Users/hbergren/Documents/xpring-eng/xpring-common-js/test/XRP/wallet.test.ts
  1:1  warning  File has too many lines (291). Maximum allowed is 250                           max-lines
```

So, if the `@see` tag gets taken care of, I either need to do a quick refactor pass to make those functions less complex / take fewer parameters, or I need to add project-specific overrides into the `xpring-common-js/.eslintrc.js` file.

keefertaylor:
Oh sweet. 

I'm pretty open to what you'd recommend as function length / parameter limits. I don't love it as an absolute heuristic, but I'm fine to try it out and see if it encourages me to have better behavior.

hbergren:
> I'm pretty open to what you'd recommend as function length / parameter limits. I don't love it as an absolute heuristic, but I'm fine to try it out and see if it encourages me to have better behavior.

Yeah, the config right now are just my best guesses. Super happy to amend them based on feedback, especially as we roll out this linting config to more projects.

In general, my thinking is:
- A function is much easier to understand if it fits on a single screen without scrolling (`line-length`)
- Cyclomatic complexity (branching, conditionals, etc) should be kept to a minimum for understandability. I think right now I have it set to `8`, which I read somewhere was a threshold for "hard to understand", but it seems [the actual standard is 10](https://github.com/eslint/eslint/issues/4808#issuecomment-167795140)
- `max-statements` usually goes along with complexity and line-length
- `max-parens` is 4 because that's a lower limit for how much state a person can hold in their head at once. Beyond that it seems reasonable to require a single object parameter, and have a TypeScript interface that defines the object.

PR #284, opened 2020-05-19
Enable code coverage configuration

keefertaylor:

Add a codecov config to Xpring-Common-JS

### Context of Change

Configure two CI checks:
- Coverage repo-wide is at least 80%
- Any single PR decreases code coverage by less than 1%

Open to discussion on numbers, these are a straw man proposal.  I'm going to leave the check as non-blocking for now - this change will only make it so we have actual numbers (otherwise code cov just compares to the existing coverage). 

I don't think this works until I have this file on master, so no UI or checks will show up here, yet. 

### Type of Change

- [ ] Documentation Updates
- [ ] Release

## Before / After

Better code!

## Test Plan

CI


amiecorso (approved):
There's some irony in this failing codecov check :) I'm assuming that's because `codecov.yml` isn't technically covered by any tests and yet counts as lines added in this PR?

stephengu (approved):
looks like some of the invalid/error cases are causing codecov to fail, did you want to have another PR add those test cases before merging this one?

keefertaylor:
TBH, I'm not sure why coverage was failing. I would expect this to be 0% change so it should pass as it now is.

PR #290, opened 2020-05-21
Small refactor 2: Code Golf

hbergren:

Mostly me trying to make some functions easier to reason about / shorter:

- Order the keys in the `TransactionJSON` object to be in the same order as the `TransactionJSON` interface
- Remove some useless `if` statements
- Refactor `additionalTransactionData` into its own helper method

### Context of Change

Just me playing code golf trying to reduce the number of exceptions to the linting config.

After this PR, I should be able to just enable `@xpring-eng/eslint-config-base` on the repo.

### Type of Change

- [ ] Documentation Updates
- [ ] Release

## Test Plan

`npm run test` is nice and green.

### DIFF
--- a/src/XRP/serializer.ts
+++ b/src/XRP/serializer.ts
@@ -54,23 +54,6 @@ const serializer = {
       SigningPubKey: '',
     }
 
-    const sequence = transaction.getSequence()?.getValue()
-    if (sequence) {
-      object.Sequence = sequence
-    }
-
-    const signingPubKeyBytes = transaction
-      .getSigningPublicKey()
-      ?.getValue_asU8()
-    if (signingPubKeyBytes) {
-      object.SigningPubKey = Utils.toHex(signingPubKeyBytes)
-    }
-
-    const lastLedgerSequence = transaction.getLastLedgerSequence()?.getValue()
-    if (lastLedgerSequence) {
-      object.LastLedgerSequence = lastLedgerSequence
-    }
-
     const normalizedAccount = getNormalizedAccount(transaction)
     if (!normalizedAccount) {
       return undefined
@@ -83,6 +66,18 @@ const serializer = {
       return undefined
     }
     object.Fee = this.xrpAmountToJSON(txFee)
+
+    // Set sequence numbers
+    object.Sequence = transaction.getSequence()?.getValue() ?? 0
+    object.LastLedgerSequence =
+      transaction.getLastLedgerSequence()?.getValue() ?? 0
+
+    const signingPubKeyBytes = transaction
+      .getSigningPublicKey()
+      ?.getValue_asU8()
+    if (signingPubKeyBytes) {
+      object.SigningPubKey = Utils.toHex(signingPubKeyBytes)
+    }
 
     const additionalTransactionData = getAdditionalTransactionData(transaction)
     if (additionalTransactionData === undefined) {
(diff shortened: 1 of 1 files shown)


src/XRP/serializer.ts
```diff
@@ -48,27 +48,10 @@ const serializer = {
...
-      .getSigningPublicKey()
-      ?.getValue_asU8()
-    if (signingPubKeyBytes) {
-      object.SigningPubKey = Utils.toHex(signingPubKeyBytes)
-    }
-
-    const lastLedgerSequence = transaction.getLastLedgerSequence()?.getValue()
-    if (lastLedgerSequence) {
-      object.LastLedgerSequence = lastLedgerSequence
+      SigningPubKey: '',
```
hbergren, line 54: Same order as the `interface BaseTransactionJSON` above. Not sure if there's a canonical order here.

src/XRP/serializer.ts
```diff
@@ -84,20 +67,23 @@ const serializer = {
...
-        }
-        Object.assign(object, this.paymentToJSON(payment))
-        break
-      }
-      default:
-        throw new Error('Unexpected transactionDataCase')
+    // Set sequence numbers
+    object.Sequence = transaction.getSequence()?.getValue() ?? 0
+    object.LastLedgerSequence =
+      transaction.getLastLedgerSequence()?.getValue() ?? 0
```
hbergren, line 73: Remove useless `if` statements here to make this shorter.

Aside: It's kind of odd that `Sequence` and `LastLedgerSequence` are both required fields, but `0` is an acceptable value if the transaction doesn't have either of those.

src/XRP/serializer.ts
```diff
@@ -84,20 +67,23 @@ const serializer = {
...
+
+    const signingPubKeyBytes = transaction
+      .getSigningPublicKey()
+      ?.getValue_asU8()
+    if (signingPubKeyBytes) {
+      object.SigningPubKey = Utils.toHex(signingPubKeyBytes)
+    }
+
+    const additionalTransactionData = getAdditionalTransactionData(transaction)
```
hbergren, line 82: made a new `getAdditionalTransactionData()` helper to make things easier to reason about.

src/XRP/serializer.ts
```diff
@@ -120,36 +106,23 @@ const serializer = {
     }
 
     // If an x-address was able to be decoded, add the components to the json.
-    const destinationAddress = payment.getDestination()
-    if (!destinationAddress) {
-      return undefined
-    }
-
-    const destination = destinationAddress.getValue()?.getAddress()
+    const destination = payment.getDestination()?.getValue()?.getAddress()
```
hbergren, line 109: Just chain the `?.` instead of having an unneeded `if` statement.

src/XRP/serializer.ts
```diff
@@ -120,36 +106,23 @@ const serializer = {
...
     if (!destination) {
       return undefined
     }
 
     const decodedXAddress = Utils.decodeXAddress(destination)
-    if (decodedXAddress) {
-      json.Destination = decodedXAddress.address
-      if (decodedXAddress.tag !== undefined) {
+    json.Destination = decodedXAddress?.address ?? destination
```
hbergren, line 115: Either it's an X-Address, or it was always a classic address (I think). Regardless of the validity of that assertion, this keeps the actual logic the same as before.

PR #294, opened 2020-05-27
Enable `@xpring-eng/eslint-config-base`

0xCLARITY:

Enable `@xpring-eng/eslint-config-base` for this repo. Part of moving all JS/TS repos over to our new config.

In particular, we:
- Upgrade `.eslintrc.js` configuration to remove unneeded config and add overrides specific to this repository where necessary
- Add various linting plugins and libraries
- Add a task for linting without automatic fixing (`lintNoFix`), which can be useful for noting violations and debugging the linting config.
- Avoid using the `object` keyword as a type
- 

### Type of Change

- [ ] Documentation Updates
- [ ] Release

## Before / After

Now this repo will be using `@xpring-eng/eslint-config-base` as its linting configuration.

## Test Plan

`npm run lintNoFix` has no linting violations.


package.json
```diff
@@ -24,24 +24,25 @@
...
     "chai": "4.2.0",
     "copyfiles": "^2.2.0",
-    "eslint": "^6.8.0",
-    "eslint-config-airbnb-base": "^14.0.0",
-    "eslint-config-airbnb-typescript": "^7.0.0",
-    "eslint-config-prettier": "^6.9.0",
+    "eslint": "^7.0.0",
     "eslint-plugin-import": "^2.20.0",
-    "eslint-plugin-mocha": "^6.2.2",
+    "eslint-plugin-jsdoc": "^25.4.2",
```
hbergren, line 42: `@xpring-eng/eslint-config-base` doesn't actually have the JSDoc rules enabled yet, but it will soon.

src/XRP/serializer.ts
```diff
@@ -7,7 +7,7 @@ import {
 } from './generated/org/xrpl/rpc/v1/transaction_pb'
 
 interface PaymentJSON {
-  Amount: object | string
+  Amount: Record<string, unknown> | string
```
hbergren, line 10: Avoid using the object type. [See reasoning here](https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/ban-types.md#default-options).

src/XRP/signer.ts
```diff
@@ -44,8 +44,8 @@ const signer = {
    * @returns A set of bytes representing the inputs and a signature.
    */
   signTransaction(
-    transaction: Transaction,
-    wallet: Wallet,
+    transaction?: Transaction,
+    wallet?: Wallet,
   ): Uint8Array | undefined {
     if (transaction === undefined || wallet === undefined) {
```
hbergren, line 50: This `if` statement means we need to make the parameters above optional, otherwise it would be "impossible" for them to be `undefined` (because TS has an unsound type system).

.eslintrc.js
```diff
@@ -19,76 +19,38 @@ module.exports = {
...
-    'plugin:prettier/recommended',
-  ],
+        // Because of gRPC, our tests have to have a ton of statements to be useful
+        'max-statements': ['warn', { max: 40 }],
 
-  rules: {
-    // This rule is about explicitly using `return undefined` when a function returns any non-undefined object.
-    // However, since we're using TypeScript, it will yell at us if a function is not allowed to return `undefined` in its signature, so we don't need this rule.
-    'consistent-return': 'off',
+        // TODO:(@hbergren) We should be able to get rid of these two rules eventually,
```
keefertaylor, line 46: nit: can you be more specific about what needs to happen before we can get rid of the rules? When can we expect this TODO to be revisited?

src/XRP/signer.ts
keefertaylor: These params shouldn't be optional, not sure why I did this. Can we remove the if statement?

keefertaylor (approved):

package.json
```diff
@@ -54,6 +56,9 @@
     "webpack": "4.43.0",
     "webpack-cli": "3.3.11"
   },
+  "engines": {
+    "node": ">=10.0.0"
+  },
```
hbergren, line 61: @keefertaylor , Does this seem right for the supported versions of Node? (`eslint-plugin-node` uses the `engines` field to determine if you're writing any code using features not supported in a version of Node you claim to support).

amiecorso (approved):
this looks good to me! congrats on final installment!

PR #295, opened 2020-05-27
Add transaction memo support in serializer

theotherian:
Signed-off-by: Ian Simpson <ian.g.simpson@gmail.com>

Add support for serializing transaction memos in order to properly sign transactions.

### Context of Change

The memo field is already part of the rippled protobuf object, so it was just a matter of serializing that to JSON in order to support signing.

### Type of Change

- [ ] Documentation Updates
- [ ] Release

## Before / After

Before: transaction memos weren't part of what was signed
After: transaction memos are added to the signature

## Test Plan

The test was updated to check for this case

### DIFF
--- a/src/XRP/serializer.ts
+++ b/src/XRP/serializer.ts
@@ -1,16 +1,23 @@
 import Utils from '../Common/utils'
 
 import { XRPDropsAmount } from './generated/org/xrpl/rpc/v1/amount_pb'
-import {
-  Payment,
-  Transaction,
-} from './generated/org/xrpl/rpc/v1/transaction_pb'
+import { Memo, Payment, Transaction, } from './generated/org/xrpl/rpc/v1/transaction_pb'
 
 interface PaymentJSON {
   Amount: object | string
   Destination: string
   DestinationTag?: number
   TransactionType: string
+}
+
+interface MemoJSON {
+  Memo?: MemoDetailsJSON
+}
+
+interface MemoDetailsJSON {
+  MemoData?: string
+  MemoType?: string
+  MemoFormat?: string
 }
 
 interface BaseTransactionJSON {
@@ -20,6 +27,7 @@ interface BaseTransactionJSON {
   Sequence: number
   SigningPubKey: string
   TxnSignature?: string
+  Memos?: MemoJSON[]
 }
 
 interface PaymentTransactionJSONAddition extends PaymentJSON {
@@ -103,6 +111,10 @@ const serializer = {
       object.TxnSignature = signature
     }
 
+    if (transaction.getMemosList() && transaction.getMemosList().length > 0) {
+      object.Memos = transaction.getMemosList().map(this.memoToJSON)
+    }
+
     return object
   },
 
@@ -162,6 +174,23 @@ const serializer = {
   xrpAmountToJSON(xrpDropsAmount: XRPDropsAmount): string {
     return `${xrpDropsAmount.getDrops()}`
   },
+
+  /**
+   * Convert a Memo to a JSON representation.
+   * @param memo - The Memo to convert.
+   * @returns The Memo as JSON.
+   */
+  memoToJSON(memo: Memo): MemoJSON {
+    const Memo: MemoDetailsJSON = {
+      MemoData: memo?.getMemoData()?.getValue_asB64(),
+      MemoFormat: memo?.getMemoFormat()?.getValue_asB64(),
+      MemoType: memo?.getMemoType()?.getValue_asB64(),
+    }
+
+    return {
+      Memo,
+    }
+  }
 }
 
 export default serializer
--- a/src/XRP/serializer.ts
+++ b/src/XRP/serializer.ts
@@ -103,13 +103,6 @@ const serializer = {
 
     Object.assign(object, this.memosToJSON(transaction.getMemosList()))
 
-    if (transaction.getMemosList() && transaction.getMemosList().length) {
-      object.Memos = transaction
-        .getMemosList()
-        .filter((memo) => !!memo)
-        .map((memo) => this.memoToJSON(memo))
-    }
-
     return object
   },
 
--- a/src/XRP/serializer.ts
+++ b/src/XRP/serializer.ts
@@ -111,9 +111,9 @@ const serializer = {
       object.TxnSignature = signature
     }
 
-    if (transaction.getMemosList() && transaction.getMemosList().length > 0) {
-      object.Memos = transaction.getMemosList().map(this.memoToJSON)
-    }
+    object.Memos = transaction
+      .getMemosList()
+      ?.map((memo) => this.memoToJSON(memo))
 
     return object
   },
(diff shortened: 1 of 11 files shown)


src/XRP/serializer.ts
```diff
@@ -135,6 +147,23 @@ const serializer = {
...
     return `${xrpDropsAmount.getDrops()}`
   },
+
+  /**
+   * Convert a Memo to a JSON representation.
+   * @param memo - The Memo to convert.
+   * @returns The Memo as JSON.
+   */
+  memoToJSON(memo: Memo): MemoJSON {
+    const Memo: MemoDetailsJSON = {
```
keefertaylor, line 157: don't uppercase your variable names

src/XRP/serializer.ts
```diff
@@ -135,6 +147,23 @@ const serializer = {
   xrpAmountToJSON(xrpDropsAmount: XRPDropsAmount): string {
     return `${xrpDropsAmount.getDrops()}`
   },
+
+  /**
+   * Convert a Memo to a JSON representation.
+   * @param memo - The Memo to convert.
```
keefertaylor, line 153: nit +1 vertical space

test/XRP/serializer.test.ts
```diff
@@ -266,4 +263,56 @@ describe('serializer', function (): void {
...
+      fee,
+      lastLedgerSequence,
+      sequence,
+      accountClassicAddress,
+      publicKey,
+    )
+
+    const memo = new Memo()
+    const memoData = new MemoData()
+    memoData.setValue('I forgot to pick up Carl...')
```
keefertaylor, line 282: You could factor these to constants, which would still be funny and then the test is marginally more maintainable at the cost of being more verbose

theotherian:
I know the formatting is off a little and I know there is a build issue. I tried using Intellij again because VSCode was acting weird about some of the git submodules and it appears I didn't get Intellij to respect the project settings.

src/XRP/serializer.ts
```diff
@@ -89,6 +97,10 @@ const serializer = {
       object.TxnSignature = signature
     }
 
+    if (transaction.getMemosList() && transaction.getMemosList().length > 0) {
```
keefertaylor, line 100: I think you can leave this off... I would imagine map() just executes 0 times.

src/XRP/serializer.ts
theotherian: I have a change for this already that I haven't committed. The reason I did this was because the variable names are upper cased in the actual spec and I was trying to reference it by name alone in the associated object.

hbergren:
> I know the formatting is off a little and I know there is a build issue. I tried using Intellij again because VSCode was acting weird about some of the git submodules and it appears I didn't get Intellij to respect the project settings.

You should be able to run `npm run lint` to find the formatting / linting violations.

theotherian:
> > I know the formatting is off a little and I know there is a build issue. I tried using Intellij again because VSCode was acting weird about some of the git submodules and it appears I didn't get Intellij to respect the project settings.
> 
> You should be able to run `npm run lint` to find the formatting / linting violations.

I know; I just foolishly thought the IDE was doing it for me and was smacked around when I ran things at the command line.

src/XRP/serializer.ts
```diff
@@ -89,6 +101,15 @@ const serializer = {
...
     }
 
+    Object.assign(object, this.memosToJSON(transaction.getMemosList()))
+
+    if (transaction.getMemosList() && transaction.getMemosList().length) {
+      object.Memos = transaction
+        .getMemosList()
+        .filter((memo) => !!memo)
+        .map((memo) => this.memoToJSON(memo))
+    }
```
hbergren, line 111: I don't understand this block of code. Why are we doing `Object.assign()` followed by directly setting `object.Memos`? What is actually happening in each of those two statements?

src/XRP/serializer.ts
```diff
@@ -135,6 +156,43 @@ const serializer = {
...
+  memoToJSON(memo: Memo): MemoJSON {
+    const jsonMemo: MemoDetailsJSON = {
+      MemoData: memo?.getMemoData()?.getValue_asB64(),
+      MemoFormat: memo?.getMemoFormat()?.getValue_asB64(),
+      MemoType: memo?.getMemoType()?.getValue_asB64(),
+    }
+
+    return {
+      Memo: jsonMemo,
+    }
```
hbergren, line 194: This feels a little odd. Isn't `jsonMemo` already a memo in JSON format? Why do we need to create a wrapper JSON object with only a `Memo` key? Is the extra layer of indirection necessary?

test/XRP/serializer.test.ts
```diff
@@ -266,4 +273,60 @@ describe('serializer', function (): void {
     }
     assert.deepEqual(serialized, expectedJSON)
   })
+
+  it('serializes a payment with a memo', function (): void {
+    // GIVEN a transaction which represents a payment to a destination without a tag, denominated in XRP, with a dank
+    // meme for a memo
```
hbergren, line 279: Dig the dank meme test.

src/XRP/serializer.ts
theotherian: That's the format of the JSON that's stored in the transaction. 

https://xrpl.org/transaction-common-fields.html#memos-field

src/XRP/serializer.ts
theotherian: I was doing it that way because the `Memo` is comprised of entirely optional fields, so I'm trying to reduce things to `undefined` if there's no actual data in there (i.e. someone who was using this passed in `Memo` instances that had nothing inside them). I was using `Object.assign` as a shorthand instead of getting the result, checking if it's not `undefined`, and then setting the field. 

I was also doing this due to our discussion the other day about the signing procedure that we have. I just figured reducing the potential footprint of the resulting JSON as much as possible was useful. 

Most of the work that is coming into question on this PR is a result of the way the rippled API is implemented for whatever that's worth.

src/XRP/serializer.ts
hbergren: Jeez. I guess I'm just unhappy with `rippled`.

amiecorso:
So this looks good to me, just one question: how is it that we were able to neglect the memo field in the past?  Is it the case that if a transaction protobuf was assembled for serialization and signing, and included a memo, that it would simply not have been included in the signed and submitted transaction?  And perhaps this wasn't an issue because the SDK doesn't currently support adding memos to the XRP transactions assembled in the `send` method?

theotherian:
@amiecorso as you said, it has more to do with the SDK not adding it in `send` than anything else. I have another branch that I'm working on which adds support for that, but this is necessary first.

src/XRP/serializer.ts
hbergren: I still don't understand why you need to do both though. Like, isn't this `.getMemosList().filter().map()` call implicit in `this.memosToJSON()`?

theotherian: Haha, I just realized I accidentally left both code paths in when I committed. Sorry; no wonder this was more confusing than necessary. Pushing a fix for that shortly.

theotherian:
@hbergren it looks like something is failing after merging in master due to the new linter settings. I'm working on trying to address it.

theotherian:
Just wanted to get the patch coverage up to 100%. Merging now.

PR #300, opened 2020-05-29
ESLint directive comments

0xCLARITY:

This updates `@xpring-eng/eslint-config-base` to `v0.5.1`. The big change is a new required plugin: `eslint-plugin-eslint-comments`.

This adds various rules around best practices for using ESLint directives, the most useful of which is requiring ESLint directive comments for `eslint-disable`s.

Feel free to review the comments I added to the `eslint-disable` directives and see if they make sense or should be reworded.

### Type of Change

- [ ] Documentation Updates
- [ ] Release

## Before / After

Now all `eslint-disable` directives have a comment explaining why the rule needs to be disabled.

## Test Plan

`npm run test` should pass as I only changed comments.


src/PayID/pay-id-components.ts
```diff
@@ -1,6 +1,8 @@
 // Disable multiple classes to accommodate the switch to idiomatic style naming.
 // TODO(keefertaylor): Remove this when migration is complete.
-/* eslint-disable  max-classes-per-file */
+/* eslint-disable  max-classes-per-file --
+ * We only export two classes in this file because one is deprecated and will eventually be removed.
+ */
```
hbergren, line 5: The basic format for an ESLint comment is just requiring two hyphens `--` followed by the comment.

You can do this all on one line, but for longer comments, I thought the multiline style I'm using here was reasonable.

keefertaylor (approved):

amiecorso (approved):
This looks good! I support enforced comments around disables.

PR #305, opened 2020-06-01
Add HD Wallet Generation to Wallet Factory

keefertaylor:

Add functions to generate a HD wallet from a seed and derivation path to `WalletFactory`. 

### Context of Change

Test cases were produced by generating a seed and derivation paths, using this well known tool: https://iancoleman.io/bip39/#english.

I then converted the classic addresses to X-addresses using xrpaddress.ino.

Note:
- This PR also refactors a method `hexFromBuffer` out of `Wallet` and onto `Utils`, which is consistent with the rest of the codebase. 
- `Wallet`'s implementation will handle invalid inputs incorrectly. 

### Type of Change

- [ ] Documentation Updates
- [ ] Release

## Before / After

New methods to generate wallet. 

## Test Plan

CI and new tests are provided.


test/XRP/wallet-factory.test.ts
```diff
@@ -79,4 +79,76 @@ describe('Wallet Factory', function (): void {
     // THEN the wallet is undefined.
     assert.isUndefined(wallet)
   })
+
+  it('walletFromSeedAndDerivationPath - valid inputs - derivation path 0', function (): void {
+    // GIVEN a wallet factory, a valid seed and a valid derivation path.
+    const walletFactory = new WalletFactory(XrplNetwork.Main)
+    const seed =
+      'f5dd847bd41be2deeb5f3596ab614a95e3cebb73dcd4270df9681de9e3ee5fe34ef07beb05fea610bf2cfdffb6dc3f1593fd966ab5afd30e803af66b68df32d4'
```
hbergren, line 87: Worth saying what this seed is not hex encoded in a comment?

0xCLARITY (approved):
Looks great! This is a solid PR.

I have some small suggestions / questions, but nothing that needs to be blocking.

Learned a bit more about how HD wallets and derivation paths work by Googling as I didn't know what was going on in this PR. Neat stuff.

amiecorso (approved):

test/XRP/wallet-factory.test.ts
keefertaylor: @hbergren I'm a bit confused what you're looking for here. Typically, what XRP Ledger refers to as a seed is a base58check encoded string beginning with 's'.  These bytes *are* hex encoded. 

Both are 'seeds' in that they are representations of bytes which are used as a seed. 

I updated this variable name to be `seedHex` to hopefully make that clearer. If you'd like, I can run this through a base58check encoder and produce a seed in base58check form. I think generally folks who are deriving HD wallets tend to be multi-coin wallets (imagine TrustWallet or Exodus) and they'll have seeds in a generic form instead of a ripple-specific form.

src/XRP/wallet-factory.ts
```diff
@@ -25,10 +31,39 @@ export default class WalletFactory {
...
+      const seedBuffer = Buffer.from(seedBytes)
+      const masterNode = bip32.fromSeed(seedBuffer)
+      const node = masterNode.derivePath(derivationPath)
+      if (node.privateKey === undefined) {
+        return undefined
+      }
+
+      const publicKey = Utils.hexFromBuffer(node.publicKey)
+      const privateKey = Utils.hexFromBuffer(node.privateKey)
+      return new Wallet(publicKey, `00${privateKey}`, this.isTest)
```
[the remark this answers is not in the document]
keefertaylor: Ultimately, we want to generate keys that can be used with `ripple-pairs`. Rippled generates private keys with a 1 byte prefix representing whether the key is secp256k1 or ed25519. The former uses bytes 0x00 and the latter uses 0xED. 

You can see these prefixes here:
https://github.com/ripple/ripple-keypairs/blob/7e67cbb9e1cc293bf696b0e0b054a214d607d28b/src/index.ts#L44

0xED is also mentioned here: https://xrpl.org/cryptographic-keys.html#ed25519-key-derivation. 

I append this prefix here because the generic libraries (bip32) don't have the concept of this seed. I could append in wallet internally, but then `walletFromSeed` which uses `ripple-keypairs` does give me this prefix this one-off logic would just move elsewhere. 

LMK what you think. After thinking about this, maybe it's "more correct" to give an unprefixed public key to the wallet and force an enum. That way when folks say "wallet.privateKey" they get a private key that works for all libraries (though not ripple-keypairs).

keefertaylor:
Re-requesting review from @hbergren  since I think he brings up some good points and want to get affirmative approval.

src/XRP/wallet-factory.ts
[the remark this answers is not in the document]
hbergren: > Rippled generates private keys with a 1 byte prefix representing whether the key is secp256k1 or ed25519. The former uses bytes 0x00 and the latter uses 0xED.

Helpful. I would at the least add that as a code comment above this line.

> Maybe it's "more correct" to give an unprefixed public key to the wallet and force an enum. That way when folks say "wallet.privateKey" they get a private key that works for all libraries.

I have no idea what usage looks like in the real world, but I would imagine that most people interacting with XRPL just use `ripple-lib` & co, right?

test/XRP/wallet-factory.test.ts
hbergren: Makes sense. Very helpful explanation.

0xCLARITY (approved):
LGTM.

I _would_ add a code comment about the `00` prefix on the private key, just for clarification purposes.

keefertaylor:
@hbergren  Added the comment. Sorry this has been a stale PR :(

PR #320, opened 2020-06-05
Add `XRPLNetwork` Enum

keefertaylor:

This PR pulls down [`XrplNetwork`](https://github.com/xpring-eng/Xpring-JS/blob/master/src/Common/xrpl-network.ts) into our common core library. 

### Context of Change

We often need to know what network components are connected to. This is best represented as an enum since there are multiple states. This enum is used liberally at the SDK library level (xpring-js, xpringkit, xpring4j) and we should use it here as well. 

Also add:
- A utility function to map a given `XrplNetwork` to whether it represents a test network
- Migrate `WalletFactory` to use this enum

The `WalletFactory` change is breaking, but no client libraries have consumed or exposed this class yet so I think it's not a big deal to change to it at this time. 

### Type of Change

- [ ] Documentation Updates
- [ ] Release

## Before / After

Networks are represented deterministically, code is more readable. 

## Test Plan

CI; additional tests provided for new methods.

## Future Tasks
- `XRPLNetwork` in Xpring-JS can be removed


src/XRP/wallet-factory.ts
```diff
@@ -1,26 +1,32 @@
...
+  public constructor(network: XrplNetwork) {
+    this.network = network
+    this.isTest = XrpUtils.isTestNetwork(network)
   }
 
   /**
    * Generate a new wallet with the given keys.
    *
-   * @param publicKey - A hexadecimal encoded public key.
+   * @param publicKey - A hexadencimal encoded public key.
```
amiecorso, line 29:
```suggestion
   * @param publicKey - A hexadecimal encoded public key.
```

src/XRP/xrpl-network.ts
```diff
@@ -0,0 +1,10 @@
+/**
+ * XRP Ledger networks components can be attached to.
```
amiecorso, line 2:
```suggestion
 * XRP Ledger networks that components can be attached to.
```

test/XRP/wallet-factory.test.ts
```diff
@@ -2,11 +2,12 @@ import { assert } from 'chai'
 
 import WalletFactory from '../../src/XRP/wallet-factory'
 import 'mocha'
+import XRPLNetwork from '../../src/XRP/xrpl-network'
```
amiecorso, line 5: should the casing here be `XrplNetwork` ?

test/XRP/wallet-factory.test.ts
```diff
@@ -2,11 +2,12 @@ import { assert } from 'chai'
...
 
 describe('Wallet Factory', function (): void {
   it('Wallet From Keys - Valid Keys', function (): void {
     // GIVEN a wallet factory and set of valid keys.
-    const walletFactory = new WalletFactory(true)
+    const walletFactory = new WalletFactory(XRPLNetwork.Test)
```
amiecorso, line 10: same for these three instances... `XrplNetwork`?

test/XRP/wallet-factory.test.ts
keefertaylor: Old habits die hard.

PR #322, opened 2020-06-08
Refactor `package.json`

0xCLARITY:

This PR does a few things:

- Run the [sort-package-json](https://www.npmjs.com/package/sort-package-json) tool to put the `package.json` keys into their canonical ordering
- Move `nyc` configuration into a standalone config file, like Prettier, ESLint, and all the other tools.

Also, I changed the color highlighting levels of NYC. Now, it will be:
- Red for < 80% coverage
- Yellow for 80% - 95% coverage
- Green for 95+% coverage

This is what I used for color highlighting in PayID, but I'm happy to change the cutoffs to align with CodeCov configuration. It _looked_ like the target was 80%, so that seemed like a reasonable lower bound.

### Context of Change

Just Hans fiddling with things.

### Type of Change

- [ ] Documentation Updates
- [ ] Release

## Test Plan

No source code touched, and `npm run test` still has the same coverage stats, just different requirements for highlighting something as green.


keefertaylor (approved):
Thank you. 

Optional: You could run the strings sorter as part of the lint process and then everything will always be sorted.

hbergren:
> Optional: You could run the strings sorter as part of the lint process and then everything will always be sorted.

Yeah, I'm thinking about the best option for that.

What I might do is take a dependency on this [ESLint plugin for package.json files](https://github.com/kellyselden/eslint-plugin-json-files#supported-rules) in `xpring-eslint`, because it gives us a few more rules.

It's a _little_ weird to be linting JSON files using a JS linter... but it allows for it, and I guess Prettier formats JSON and HTML and `.vue` and everything else, so it's not that weird for a JS build tool.

amiecorso (approved):

PR #329, opened 2020-06-09
Support AccountSet Operations in Serialization

keefertaylor:

Add support for AccountSet Serialization.

### Context of Change

We'd like to begin to support additional operation types. `Serializer` acts as a shim layer that converts protocol buffers to JSON objects. We can then feed those JSON objects to ripple-binary-codec and get bytes to sign.

### Type of Change

- [ ] Documentation Updates
- [ ] Release

## Before / After

Additional serialization is supported. 

## Test Plan

New tests added for CI.


test/XRP/serializer.test.ts
```diff
@@ -164,6 +174,127 @@ function makeDepositPreauth(
...
+    senderAddress,
+    publicKey,
+  )
+  transaction.setAccountSet(accountSet)
+
+  return transaction
+}
+
+/**
+ * Make an AccountSet protocol buffer fro the given inputs.
```
sarahkevinking, line 230: nit: `for`

keefertaylor: Done. I used `from` instead of `for`. LMK if that distinction bugs you.

src/XRP/serializer.ts
```diff
@@ -1,15 +1,28 @@
+/* eslint-disable  max-lines --
+ * gRPC is verbose. Playing code golf wiht this file would decrease clarity for little readability gain.
```
sarahkevinking, line 2: nit: with

Code golf sounds like a fun game

src/XRP/serializer.ts
```diff
@@ -1,15 +1,28 @@
+/* eslint-disable  max-lines --
```
tedkalaw, line 1: is `max-lines` too strict?

keefertaylor:
> did we want to have any kind of basically number -> uint8 (ex: for TickSize) checks anywhere? let me know if it's being done somewhere else

Not positive of what you're asking.  Do you mean making sure the number is 3..15? 
Screen Shot 2020-06-24 at 4 41 12 PM

stephengu:
> > did we want to have any kind of basically number -> uint8 (ex: for TickSize) checks anywhere? let me know if it's being done somewhere else
> 
> Not positive of what you're asking. Do you mean making sure the number is 3..15?
> Screen Shot 2020-06-24 at 4 41 12 PM

yes exactly

amiecorso:
I discussed with Keefer and in keeping with how we've handled this type of logic elsewhere, we're leaving sanity checks to the SDK layer that actually constructs or handles the protobufs, instead of the serialization code.

PR #357, opened 2020-06-29
Update @xpring/eslint

0xCLARITY:

Update `@xpring-eng/eslint-config-base` and related peer dependencies.

### Context of Change

I just cut a new version of `@xpring/eslint`, so I figured I would update it in here. I think it's not auto-updating with Dependabot because it's pre-1.0, and thus could (and sometimes does) have breaking changes in every version.

### Type of Change

- [ ] Documentation Updates
- [ ] Release

## Test Plan

`npm run lintNoFix` has no linting errors.


src/XRP/serializer.ts
```diff
@@ -271,7 +271,10 @@ function getNormalizedAccount(transaction: Transaction): string | undefined {
  * Given a Transaction, get the JSON representation of data specific to that transaction type.
  *
  * @param transaction - A transaction to check for additional transaction-type specific data.
+ *
  * @returns A JSON representation of the transaction-type specific data, or undefined.
+ *
+ * @throws An error if given a transaction that we do not know how to handle.
```
hbergren, line 277: We now require documenting `throws` in JSDoc if the function throws an error.

src/XRP/wallet-factory.ts
```diff
@@ -38,6 +38,7 @@ export default class WalletFactory {
    * @param derivationPath - The given derivation path to use. If undefined, the default path is used.
    * @returns A new wallet from the given mnemonic if the mnemonic was valid, otherwise undefined.
    */
+  // eslint-disable-next-line max-statements -- This does not read as too complex a function.
   public walletFromSeedAndDerivationPath(
```
hbergren, line 42: This function is the one that logs to the console, which is why we disable that rule in `.eslintrc.js`.

src/XRP/wallet-factory.ts
```diff
@@ -38,6 +38,7 @@ export default class WalletFactory {
    * @param derivationPath - The given derivation path to use. If undefined, the default path is used.
    * @returns A new wallet from the given mnemonic if the mnemonic was valid, otherwise undefined.
    */
+  // eslint-disable-next-line max-statements -- This does not read as too complex a function.
```
hbergren, line 41: We have a limit of 10 statements per function.

Possibly the following could actually be pulled out into its own function like `getBip32Node()`, and then we wouldn't need this disable?

```ts
const seedBytes = Utils.toBytes(seed)
const seedBuffer = Buffer.from(seedBytes)
const masterNode = bip32.fromSeed(seedBuffer)
const node = masterNode.derivePath(derivationPath)
if (node.privateKey === undefined) {
  return undefined
}
```

package.json
```diff
@@ -29,19 +29,21 @@
...
     "chai": "4.2.0",
     "copyfiles": "^2.2.0",
-    "eslint": "^7.0.0",
+    "eslint": "^7.3.1",
+    "eslint-plugin-array-func": "^3.1.6",
     "eslint-plugin-eslint-comments": "^3.2.0",
-    "eslint-plugin-import": "^2.20.0",
-    "eslint-plugin-jsdoc": "^27.0.0",
+    "eslint-plugin-import": "^2.22.0",
+    "eslint-plugin-jsdoc": "^28.5.1",
```
hbergren, line 46: Short summary:
`chai-as-promised`: Tricky extension for Chai that makes Chai/Mocha work better with Promises... but things break if you don't `return` the `chai-as-promsied` assertions from your tests, so this adds linting rules for that.

`eslint-plugin-array-func`: Make sure you're using `Array.from()` and various other array functions correctly.

`eslint-plugin-import 2.22.0`: Waaaay better support for TypeScript and types/interfaces.

`eslint-plugin-jsdoc`: Adds a rule requiring documenting errors with the JSDoc `@throws` keyword.

amiecorso (approved):

keefertaylor (approved):

PR #408, opened 2020-08-01
Serialization for IssuedCurrencyAmounts

keefertaylor:

Add support for serializing IssuedCurrencyAmounts.

### Context of Change

We currently don't serialize all fields in transactions. This is a step towards fixing that. 

### Type of Change

- [ ] Documentation Updates
- [ ] Release

## Before / After

No change since this isn't hooked up to anything yet. Future PRs will build on this to add in support for Paths in payments. 

## Test Plan

CI; new tests provided.

### DIFF
--- a/src/XRP/serializer.ts
+++ b/src/XRP/serializer.ts
@@ -6,7 +6,7 @@ import Utils from '../Common/utils'
 import {
   XRPDropsAmount,
   Currency,
-  IssuedCurrencyAmount
+  IssuedCurrencyAmount,
 } from './generated/org/xrpl/rpc/v1/amount_pb'
 import {
   AccountSet,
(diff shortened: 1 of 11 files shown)


src/XRP/serializer.ts
```diff
@@ -366,6 +376,34 @@ const serializer = {
...
+   * @param issuedCurrencyAmount - The {@link IssuedCurrencyAmount} to convert.
+   * @returns A JSON representation of the input.
+   */
+  issuedCurrencyAmountToJSON(
+    issuedCurrencyAmount: IssuedCurrencyAmount,
+  ): IssuedCurrencyAmountJSON | undefined {
+    const currencyWrapper = issuedCurrencyAmount.getCurrency()
+    const value = issuedCurrencyAmount.getValue()
+    const issuer = issuedCurrencyAmount.getIssuer()?.getAddress()
+    if (currencyWrapper === undefined || value === '' || issuer === undefined) {
```
amiecorso, line 391: are you sure that a missing `value` entry in an `IssuedCurrencyAmount` manifests as an empty string? Might be worth one more test that populates that the `currency` and `issuer` fields that hopefully still returns undefined, but if you're sure about it the logic seems obvious enough to me.

keefertaylor: Sure. Done.

PR #409, opened 2020-08-01
Serialization for Currency

keefertaylor:

Add support for serializing Currency objects.

### Context of Change

We currently don't serialize all fields in transactions. This is a step towards fixing that. 

Currencies are serialized as either a hex encoded string or a currency string. See: https://xrpl.org/currency-formats.html#currency-codes

### Type of Change

- [ ] Documentation Updates
- [ ] Release

## Before / After

No change since this isn't hooked up to anything yet. Future PRs will build on this to add in support for Paths in payments. 

## Test Plan

CI; new tests provided.

### DIFF
--- a/src/XRP/serializer.ts
+++ b/src/XRP/serializer.ts
@@ -3,7 +3,7 @@
  */
 import Utils from '../Common/utils'
 
-import { XRPDropsAmount } from './generated/org/xrpl/rpc/v1/amount_pb'
+import { XRPDropsAmount, Currency } from './generated/org/xrpl/rpc/v1/amount_pb'
 import {
   AccountSet,
   Memo,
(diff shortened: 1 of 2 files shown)


src/XRP/serializer.ts
```diff
@@ -77,6 +77,7 @@ interface PathElementJSON {
   currencyCode?: string
 }
 
+type CurrencyJSON = string
```
amiecorso, line 80: does a naked string count as JSON?  news to me, but seems reasonable

amiecorso (approved):

src/XRP/serializer.ts
Stormtv: Yeah, I think this type is unnecessary since the code referencing this is returning hex in a string format.

Stormtv (approved):
Looks good but I would recommend removing the CurrencyJSON type and using string

keefertaylor:
I'd prefer to have this stuff typed for safety / readability.

PR #419, opened 2020-08-05
Implement Serialization For AccountAddress

keefertaylor:

Add Serialization for AccountAddresses.

### Context of Change

This was not implemented previously because ripple-binary codec did not support X-Addresses. In the most recent release this is true however. Release is currently `next`, which I believe is safe to upgrade to. 

Accepting this PR means that we can begin to clean up edge cases around `AccountAddress` in `Serializer` and have much cleaner logic which will benefit us in the long term. I believe the benefits greatly outweigh the risk of using a cutting edge version of ripple-binary-codec, however, reviewers may have a different opinion and we can delay this PR. 

### Type of Change

- [ ] Documentation Updates
- [ ] Release

## Before / After

N/A

## Test Plan

New unit tests provided. Existing tests on CI prove that this change is a no-op.


keefertaylor:
Taking a pretty large reviewer list here since this change does upgrade us to a bleeding edge ripple-binary-codec and I want to gain wide spread consensus that this is the right move. Thanks in advance for spirited discussion and any insights you can provide.

tedkalaw (approved):
seems reasonable to me to use the new rbc! nice

amiecorso (approved):

intelliot (approved):
I've completed some additional (manual) testing of ripple-binary-codec 1.0.0-rc3 and as far as I can tell, everything works 👍

PR #421, opened 2020-08-05
Refactor Unauthorization Serialization to a helper function.

keefertaylor:

Refactors the code to serialize an Unauthorize protocol buffer to a helper method. 

Refactors some common code to be de-duplicated. 

### Context of Change

### Type of Change

- [ ] Documentation Updates
- [ ] Release

## Before / After

N/A

## Test Plan

CI

### DIFF
--- a/src/XRP/serializer.ts
+++ b/src/XRP/serializer.ts
@@ -4,6 +4,7 @@
 import Utils from '../Common/utils'
 
 import { XRPDropsAmount, Currency } from './generated/org/xrpl/rpc/v1/amount_pb'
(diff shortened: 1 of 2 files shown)


src/XRP/serializer.ts
```diff
@@ -77,6 +78,7 @@ interface PathElementJSON {
   currencyCode?: string
 }
 
+type UnauthorizeJSON = string
```
tedkalaw, line 81: will this eventually be an object-y type? although i suppose that a string is valid JSON

PR #423, opened 2020-08-06
Add Helper Function for ClearFlag Serialization #422

keefertaylor:

Add function for `ClearFlag` serialization.

### Context of Change

`ClearFlag` docs: https://xrpl.org/accountset.html

### Type of Change

- [ ] Documentation Updates
- [ ] Release

## Before / After

N/A

## Test Plan

New tests provided for CI.

### DIFF
--- a/test/XRP/serializer.test.ts
+++ b/test/XRP/serializer.test.ts
@@ -926,7 +926,7 @@ describe('serializer', function (): void {
   })
 
   it('Serializes a ClearFlag', function (): void {
-    // GIVEN a ClearFlag with no fields set.
+    // GIVEN a ClearFlag.
     const flagValues = 1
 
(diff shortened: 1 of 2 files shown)


test/XRP/serializer.test.ts
```diff
@@ -923,4 +924,18 @@ describe('serializer', function (): void {
     // WHEN it is serialized THEN the result is undefined.
     assert.isUndefined(Serializer.currencyToJSON(currency))
   })
+
+  it('Serializes a ClearFlag', function (): void {
+    // GIVEN a ClearFlag with no fields set.
```
amiecorso, line 929: doesn't this have one field set?  the `value` ?

keefertaylor: Done

PR #425, opened 2020-08-06
Add Helper Function for EmailHash Serialization

keefertaylor:

Add function for `EmailHash` serialization.

### Context of Change

`EmailHash` docs: https://xrpl.org/accountset.html

### Type of Change

- [ ] Documentation Updates
- [ ] Release

## Before / After

N/A

## Test Plan

New tests provided for CI.

### DIFF
--- a/test/XRP/serializer.test.ts
+++ b/test/XRP/serializer.test.ts
@@ -888,7 +888,7 @@ describe('serializer', function (): void {
     assert.deepEqual(serialized[0], Serializer.pathElementToJSON(pathElement1))
     assert.deepEqual(serialized[1], Serializer.pathElementToJSON(pathElement2))
   })
-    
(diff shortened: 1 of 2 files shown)


test/XRP/serializer.test.ts
```diff
@@ -923,4 +923,18 @@ describe('serializer', function (): void {
...
   })
+
+  it('Serializes an EmailHash', function (): void {
+    // GIVEN an EmailHash.
+    const emailHashBytes = new Uint8Array([1, 2, 3, 4])
+
+    const emailHash = new EmailHash()
+    emailHash.setValue(emailHashBytes)
+
+    // WHEN it is serialized
```
tedkalaw, line 934: super nit: `.`

PR #442, opened 2020-08-12
Pull in @payid-org/utils for PayID parsing

keefertaylor:
## High Level Overview of Change

Use official PayId library for parsing. 

### Context of Change

This change simply defers the finer points of PayID parsing to the PayID libraries for consistency. Our PayID tests still pass, proving that this method's behavior did not change. 

Also deprecate a hanging method that we no longer use. 

### Type of Change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [x] Refactor (non-breaking change that only restructures code)
- [ ] Tests (You added tests for code that already exists, or your new feature included in this PR)
- [ ] Documentation Updates
- [ ] Release

## Before / After

N/A

## Test Plan

CI

PR #452, opened 2020-08-18
Revert #442

keefertaylor:

Revert #442 which introduced an unpackable payid parser

### Context of Change

After merging #442, we cannot execute this code in Swift / Java VMs.

### Type of Change

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] Documentation Updates
- [ ] Release

## Before / After

Builds should work

## Test Plan

CI

### DIFF
--- a/src/PayID/pay-id-utils.ts
+++ b/src/PayID/pay-id-utils.ts
@@ -1,5 +1,3 @@
-import * as utils from '@payid-org/utils'
-
 import PayIdComponents from './pay-id-components'
 
 /**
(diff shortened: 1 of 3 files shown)


tedkalaw (approved):
seems ok

what error does `webpack` give? assuming that there's something going on w `crypto` or another `node`y thing

src/PayID/pay-id-utils.ts
```diff
@@ -13,13 +11,38 @@ const payIdUtils = {
...
+  /**
+   * Validate if the input is ASCII based text.
+   *
+   * Shamelessly taken from:
+   * https://stackoverflow.com/questions/14313183/javascript-regex-how-do-i-check-if-the-string-is-ascii-only.
+   *
+   * @param input - The input to check.
+   * @returns A boolean indicating the result.
+   */
+  isASCII(input: string): boolean {
```
hbergren, line 43: It's not clear to me that we need ASCII, since in most languages, requesting a unicode URL will automatically be transcribed into the underlying ASCII representation, so in that sense, unicode PayIDs are potentially valid.

0xCLARITY (approved):
Approving, with one comment.

src/PayID/pay-id-utils.ts
keefertaylor: Done.

amiecorso (approved):

PR #485, opened 2020-08-31
Serialization for Condition

tedkalaw:

Provides serialization for Condition protocol buffers. 

### Context of Change

Each protocol buffer (`Foo`) maps to an equivalent `FooJSON` in `Serializer`. This PR wires this conversion for `Condition` and adds associated unit tests. 

### Type of Change

- [ ] Documentation Updates
- [ ] Release

## Before / After

N/A

## Test Plan

CI - new tests provided.

### DIFF
--- a/src/XRP/serializer.ts
+++ b/src/XRP/serializer.ts
@@ -833,6 +833,7 @@ const serializer = {
     }
     return this.accountAddressToJSON(accountAddress)
   },
+
   /**
    * Convert a CheckID to a JSON representation.
    *
--- a/src/XRP/serializer.ts
+++ b/src/XRP/serializer.ts
@@ -37,6 +37,7 @@ import {
   Account,
   OfferSequence,
   Owner,
+  Condition,
 } from './generated/org/xrpl/rpc/v1/common_pb'
 import {
   AccountSet,
@@ -182,6 +183,7 @@ type PathJSON = PathElementJSON[]
 type CurrencyJSON = string
 type OfferSequenceJSON = number
 type OwnerJSON = string
+type ConditionJSON = string
 
 /**
  * Provides functionality to serialize from protocol buffers to JSON objects.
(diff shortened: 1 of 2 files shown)


src/XRP/serializer.ts
```diff
@@ -917,6 +919,16 @@ const serializer = {
...
   },
+
+  /**
+   * Convert a Condition to a JSON representation.
+   *
+   * @param condition - The Condition to convert.
+   * @returns The Condition as JSON.
+   */
+  conditionToJSON(condition: Condition): ConditionJSON {
+    return condition.getValue_asB64()
```
keefertaylor, line 930: This will give you a base 64 encoded string. I think that it's meant to be encoded as hexadecimal: https://xrpl.org/escrow-object.html

You can use `Utils.toHex` and `getValue_asU8()` to get a hex encoded string. 

Aside: I think it would be okay to return a UInt8Array from this method but in other places we've opted to return a hex-encoded string so I would value consistency.

test/XRP/serializer.test.ts
```diff
@@ -1586,4 +1587,18 @@ describe('serializer', function (): void {
     // THEN the result is undefined.
     assert.isUndefined(serialized)
   })
+
+  it('Serializes a Condition', function (): void {
+    // GIVEN a Condition.
+    const conditionByteString =
+      'A0258020E3B0C44298FC1C149AFBF4C8996FB92427AE41E4649B934CA495991B7852B855810100'
+    const condition = new Condition()
+    condition.setValue(conditionByteString)
```
keefertaylor, line 1596: I *think* your hex encoded string is being interpreted as base64 which is why this works.

src/XRP/serializer.ts
tedkalaw: sweet, this is the exact question i was gonna ask. will fix this

PR #510, opened 2020-09-01
Serializes SetRegularKey

tedkalaw:

Serialize SetRegularKey

### Context of Change

Adds serialization for SetRegularKey

Docs: https://xrpl.org/setregularkey.html

### Type of Change

- [ ] Documentation Updates
- [ ] Release

## Before / After

N/A

## Test Plan

CI - Tests updated

### DIFF
--- a/src/XRP/serializer.ts
+++ b/src/XRP/serializer.ts
@@ -1301,7 +1301,7 @@ const serializer = {
    * @param setRegularKey - The SetRegularKey to convert.
    * @returns The SetRegularKey as JSON.
    */
-  setRegularKeyToJSON(setRegularKey: SetRegularKey): SetRegularKeyJSON {
+  setRegularKeyToJSON(setRegularKey: SetRegularKey): SetRegularKeyJSON | undefined {
     const json: SetRegularKeyJSON = {
       TransactionType: 'SetRegularKey',
     }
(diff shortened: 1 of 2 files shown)


src/XRP/serializer.ts
```diff
@@ -1285,6 +1294,31 @@ const serializer = {
...
+  ): SetRegularKeyJSON | undefined {
+    const json: SetRegularKeyJSON = {
+      TransactionType: 'SetRegularKey',
+    }
+
+    const regularKey = setRegularKey.getRegularKey()
+    if (regularKey) {
+      const regularKeyJson = this.regularKeyToJSON(regularKey)
+      if (!regularKeyJson) {
+        return undefined
```
tedkalaw, line 1315: let me know if you think this is too defensive

amiecorso: I don't think this is too defensive:  I actually made a note of this pattern as something that we should probably address universally... because in many cases around optional fields we check whether the protobuf has the optional field, and then we just go ahead and convert it to JSON and use the result... but if the protobuf has a real field then we should definitely be throwing or something if the conversion to JSON fails... we don't want it to fail silently.  We might want to audit this all at once.

keefertaylor: I personally would not have this check, so at least the file is consistent in its behavior. 

At the end of the day, if you pass in something malformed you're going to get something malformed out. We can try our best to let that not happen, but I think the existing pattern is a reasonable trade off.

PR #522, opened 2020-09-03
Serialize Transaction Data for all Transaction Types

keefertaylor:

Serialize transaction specific data for all functions now that serialization for these protocol buffers are implemented.

### Context of Change

Note that this change is just hooking up protocol buffers. 

The protocol buffer library ensures (via a `oneof`) that if:
```
transaction.getTransactionDataCase() === Transaction.TransactionDataCase.FOO
```

then 
```
transaction.getFoo() !== undefined
```

For that reason, lots of lines in this PR are not tested because they are a logical fallacy based on the sanity checks provided by generated protocol buffer code.

Also note that all transaction data cases are not hit because we know:
- Every subpayment type, ex. `bar`, serializes correctly (via unit tests for `barToJSON`)
- Top level payment fields serialize correctly, via existing unit tests (and soon via improved tests in #519)

### Type of Change

- [ ] Documentation Updates
- [ ] Release

## Before / After

After this change, we can serialize all specific transaction data. Note that some common fields, such as `signers` are still not serialized. That functionality is coming in #519. 

Also note there are several TODOs in this class which will get cleaned up as a final task for this functionality. 

## Test Plan

CI shows that this does not affect existing behavior. Rational for not providing additional unit tests is provided above.


amiecorso (approved):
awesome

tedkalaw (approved):
nice, seems reasonable

PR #523, opened 2020-09-03
Refactors Test Utilities, Adds Signing Test for All Payment Fields

amiecorso:
This PR refactors the test utility functions into a separate helper file for re-use, and then uses these helpers in the process of a new `signer` test to test the signing of a `Payment` transaction with ALL fields set (now that the serializer supports serialization of all fields).

### Context of Change
With the new version of `xpring-common-js` the SDKs tests were suddenly failing signing.  This `signer` test illuminates that some of the newly serialized `Payment` fields encounter strict format enforcements in the `ripple-binary-codec`.  Therefore, our integration tests in the SDKs need to use test values that conform to these realistic format requirements.

### Type of Change

## Before / After
Helper utilities refactored for better organization and re-usability.
New test for `signer` sanity-checks signing of the newly expanded `Payment` JSON.

## Test Plan
New test.  CI passes. 

Update the values used in integration tests in the SDKs to be in realistic, valid formats for consumption by the ripple binary codec.

### DIFF
--- a/test/XRP/helpers/xrp-test-utils.ts
+++ b/test/XRP/helpers/xrp-test-utils.ts
@@ -29,13 +29,7 @@ import {
   AccountSet,
 } from '../../../src/XRP/generated/org/xrpl/rpc/v1/transaction_pb'
 
-
 const xrpTestUtils = {
-
-  /* eslint-disable no-shadow, max-params --
-  * The values we are shadowing are only used as inputs for this function,
-  * and it's fine to have a ton of parameters because this function is only used for testing purposes.
-  */
   /**
    * Create a new `Transaction` object with the given inputs.
    *
@@ -315,7 +309,6 @@ const xrpTestUtils = {
 
     return transaction
   },
-  /* eslint-enable no-shadow, max-params */
 
   /**
    * Make a PathElement.
@@ -331,7 +324,7 @@ const xrpTestUtils = {
     account: AccountAddress | undefined,
     currencyCode: Uint8Array | undefined,
     issuer: AccountAddress | undefined,
-  ) {
+  ): Payment.PathElement {
     const pathElement = new Payment.PathElement()
 
     if (account !== undefined) {
@@ -357,7 +350,7 @@ const xrpTestUtils = {
    * @param drops - A numeric string representing the number of drops.
    * @returns A new XRPDropsAmount.
    */
-  makeXrpDropsAmount(drops: string) {
+  makeXrpDropsAmount(drops: string): XRPDropsAmount {
     const xrpDropsAmount = new XRPDropsAmount()
     xrpDropsAmount.setDrops(drops)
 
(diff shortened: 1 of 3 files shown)


test/XRP/serializer.test.ts
```diff
@@ -1,6 +1,9 @@
+/* eslint-disable max-statements --
```
amiecorso, line 1: This should get rid of that ubiquitous yellow highlighting in serializer test

test/XRP/signer.test.ts
```diff
@@ -95,8 +101,124 @@ describe('signer', function (): void {
...
+    const wallet = new FakeWallet(fakeSignature)
+
+    const value = '1000'
+    const destinationAddress = 'XVPcpSm47b1CZkf5AkKM9a84dQHe3m4sBhsrA4XtnBECTAc'
+    const fee = '10'
+    const sequenceNumber = 1
+    const account = 'X7vjQVCddnQ7GCESYnYR3EdpzbcoAMbPw7s2xv8YQs94tv4'
+    const destinationTagValue = 4
+    // invoiceId value should be either a base64 string or a 32-byte array representing 64 hex chars (256 bits)
+    const invoiceIdValue = 'ungWv48Bz+pBQUDeXa4iI7ADYaOWF3qctBD/YfIAFa0='
```
amiecorso, line 116: this needs to be in a valid format (see comment above)

test/XRP/helpers/xrp-test-utils.ts
```diff
@@ -0,0 +1,410 @@
...
+  TickSize,
+} from '../../../src/XRP/generated/org/xrpl/rpc/v1/common_pb'
+import {
+  Payment,
+  Transaction,
+  DepositPreauth,
+  AccountSet,
+} from '../../../src/XRP/generated/org/xrpl/rpc/v1/transaction_pb'
+
+const xrpTestUtils = {
```
tedkalaw, line 32: this seems fine but in ES6, native modules are objects key'd by the symbol they're exporting. there are a few ways you can do this:

```
function a() {}
function b() {}

export default { a, b }
```

since the importer names the import, this is equivalent to what you have right now and then doing `import xrpTestUtils from 'pathtothisfile'`. the downside of this is that you must explicitly add it to the exported object, but i feel it's a bit easier to see what things are being exported by name

the default export is always named by the exporter (you can think of it as a doing a `let xrpTestUtils = module.default` thus, you could do something `import myBoguslyNamedFunction from 'pathtothisfile`

if you want, i think it's also fine to just not have a default export. if you remove the default export but export all of the functions in this file, at the import site you can do

```
import * as xrpTestUtils from 'pathtothisfile'
```

the second is my preference but the sdk seems to prefer explicit namespaced exports. it lets you import only the functions you want but you lose the explicit `xrpTestUtils` token. up to you

tedkalaw (approved):

test/XRP/helpers/xrp-test-utils.ts
amiecorso: This makes sense for the most part.  For now, and since it's test code, I'm going to leave it in the same style as the other modules in `x-c-js` - but in future will favor the latter.

natenichols (approved):

PR #528, opened 2020-09-10
Refactor Memo List Serialization

keefertaylor:

Refactor memo list to be standardized. 

### Context of Change

Refactor method to use same idioms as other list serialization methods (see for example, `pathListToJSON`). 

Specifically:
- Change name to `memoListToJSON` for consistency
- Use a simple map function internally
- Do not check inputs for validity, defer that responsibility to callers (note, this removes a unit test)
- Input is `Array<Memo>` and output is `Array<MemoJSON`>

### Type of Change

- [ ] Documentation Updates
- [ ] Release

## Before / After

No change. This is prep work for a larger refactor of common transaction fields in #519 

## Test Plan

CI shows that this does not affect the public API; removed a unit test which was testing removed functionality.

### DIFF
--- a/src/XRP/serializer.ts
+++ b/src/XRP/serializer.ts
@@ -436,7 +436,10 @@ const serializer = {
       object.TxnSignature = signature
     }
 
-    Object.assign(object, this.memosToJSON(transaction.getMemosList()))
+    const memoList = transaction.getMemosList()
+    if (memoList.length > 0) {
+      object.Memos = this.memoListToJSON(memoList)
+    }
 
     const additionalTransactionData = getAdditionalTransactionData(transaction)
     if (additionalTransactionData === undefined) {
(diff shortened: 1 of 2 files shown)


src/XRP/serializer.ts
```diff
@@ -805,13 +808,9 @@ const serializer = {
...
-  memosToJSON(memos: Memo[]): { Memos: MemoJSON[] } | undefined {
-    if (!memos.length) {
-      return undefined
-    }
-
-    const convertedMemos = memos.map((memo) => this.memoToJSON(memo))
-    return { Memos: convertedMemos }
+  memoListToJSON(memos: Memo[]): MemoJSON[] {
+    // eslint-disable-next-line @typescript-eslint/unbound-method -- Manually assigning `this`.
+    return memos.map(this.memoToJSON, this)
```
tedkalaw, line 813: can you use an arrow function instead here?

```
return memos.map((memo) => this.memoToJSON(memo))
```

`this` is presered in arrow functions

keefertaylor: Totally. I was trying to make the linter happy here (we do this other places too). 

Do we have any thoughts on whether we should disable this linter rule?

keefertaylor:
/+ Hans who may have comments about the linter. I think you've mentioned we could disable the rule in question in the past, I just forget exactly what you've said :)

keefertaylor:
Gentle ping, @tedkalaw

amiecorso (approved):

src/XRP/serializer.ts
tedkalaw: it gets mad with arrow functions? the dox for the rule implies that arrow funcs are cool: https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/unbound-method.md

if we're doing this other places already, i say just go for it as is

tedkalaw (approved):

PR #542, opened 2020-09-21
Uses @payid-org/utils for PayID parsing

tedkalaw:
This re-enables the use of `@payid-org/utils` for parsing PayIDs. Previously, the build wasn't working because `@payid-org/utils` is using the `fs` module for two functions; since there is no filesystem available in the browser, these functions wouldn't be used there anyway.

### Context of Change
Originally added in https://github.com/xpring-eng/xpring-common-js/pull/442 and reverted in https://github.com/xpring-eng/xpring-common-js/pull/452. We decided we're ok with the described issues in Swift and JVM, as we have an alternative fix.

### Type of Change

- [ ] Documentation Updates
- [ ] Release

## Before / After
Parsing PayIDs uses @payid-org/utils

## Test Plan
CI

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -22,6 +22,10 @@ module.exports = {
   resolve: {
     extensions: ['.ts', '.js'],
   },
+  node: {
+    // @payid-org/utils has a dependency on fs
+    fs: 'empty',
+  },
   output: {
     filename: 'index.js',
     path: path.resolve(__dirname, 'dist'),
--- a/package-lock.json
+++ b/package-lock.json
@@ -374,6 +374,21 @@
         }
       }
     },
+    "@panva/asn1.js": {
+      "version": "1.0.0",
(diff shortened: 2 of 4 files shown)


webpack.config.js
```diff
@@ -22,6 +22,10 @@ module.exports = {
   resolve: {
     extensions: ['.ts', '.js'],
   },
+  node: {
+    // @payid-org/utils has a dependency on fs
+    fs: 'empty',
```
tedkalaw, line 27: let me know if you guys think this is bogus; this is used specifically in https://github.com/payid-org/utils/blob/master/src/verifiable/keys.ts#L17. the functions that operate on files can never work in the browser, so this feels alright to me, although perhaps i should update the documentation...

amiecorso (approved):

webpack.config.js
nhartner: @hbergren is working on removing that dependency because it's only needed by the PayID cli. This looks like an acceptable work around until then.

nhartner (approved):

PR #550, opened 2020-10-05
allow builds targeting node

ledhed2222:

Allow builds targeting node instead of web. This will allow the Ruby SDK to take advantage of latest version of this library.

### Context of Change

The Ruby SDK requires use of this library via node. The current library only will build with web targets, making it not possible to import via node `require` syntax.

### Type of Change

- [ ] Documentation Updates
- [ ] Release

## Before / After

Gives a new command line option from `package.json` called `webpack-node` that will build with a node target.

## Test Plan

Built the new version successfully and was able to import libraries via node using it.


tedkalaw (approved):
cool, seems reasonable. we may have to revisit later as we work out webpacky things and xcj generally, but gogogo

mvadari (approved):

PR #555, opened 2020-10-06
Serializes TrustSet Transactions

mvadari:

Provides serialization of TrustSet transaction protocol buffers. 

### Context of Change

`TrustSet` is one of the few transaction types in use at the moment. It was unable to be properly serialized. 

This was caught in the process of writing more `Signer` tests for transactions.

### Type of Change

## Before / After

`TrustSet` transactions can now be fully serialized. 

## Test Plan

CI passes. New unit test written.

### DIFF
--- a/src/XRP/serializer.ts
+++ b/src/XRP/serializer.ts
@@ -251,6 +251,7 @@ type TransactionDataJSON =
   | PaymentChannelCreateJSON
   | PaymentChannelFundJSON
   | SetRegularKeyJSON
+  | TrustSetJSON
 
 /**
  * Individual Transaction Types.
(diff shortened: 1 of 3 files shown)


tedkalaw (approved):
looks good

amiecorso (approved):
yeah this looks good to me too, not sure what codecov is on about.  probably because we're not testing our test helpers or whatever, so I don't think we need to worry about the 50% diff coverage

mvadari:
> not sure what codecov is on about

I think it's because I only added like 4 lines of codecov-checking code, and that happens to include an undefined check that can't really be triggered (but Typescript doesn't know that).

PR #556, opened 2020-10-07
Adds TrustSet Transaction Tests for `Signer`

mvadari:

This PR adds additional testing for `Signer` for `TrustSet` transactions. 

### Context of Change

This is a part of an initiative to have better testing for `Signer`, to catch bugs quickly and more easily in the future. 

### Type of Change

## Before / After

New tests added for `Signer`. 

## Test Plan

CI passes. 

## Future Tasks
More tests for `Signer` will be added.


test/XRP/fakes/fake-xrp-protobufs.ts
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable max-statements -- long functions are fine here */
```
amiecorso, line 1: This is a weird request, but would you mind renaming this file to "fake-protobufs.ts" without the "xrp" ? I know it's probably called this in `xpring-js`, but I feel like our indiscriminate use of "xrp" in naming things is a relic of when the design of the SDK was going to be limited to only XRP-related functionality.  In most places it's used, we should really have been using either "Xrpl" to indicate the XRP Ledger, or nothing at all.  So just a philosophy we should think about going forward.

test/XRP/fakes/fake-xrp-protobufs.ts
```diff
@@ -284,68 +309,118 @@ paymentAllFields.setDeliverMin(deliverMin)
...
+ * @param transactionType - The type of transaction that is created.
+ * @param object - The object to be inserted into the transaction based on the transaction type.
  * @returns Payment Transaction with the included payment param.
+ * @throws Error if given bad data.
  */
 function buildStandardTestTransaction(
-  accountSet?: AccountSet,
-  payment?: Payment,
+  transactionType: Transaction.TransactionDataCase,
+  object: AccountSet | Payment | TrustSet,
```
amiecorso, line 339: Nice, this'll scale. maybe `transactionData` or `additionalFields` would be a slightly more descriptive param

test/XRP/signer.test.ts
```diff
@@ -345,4 +352,102 @@ describe('Signer', function (): void {
...
+      wallet,
+    )
+
+    // THEN the signing artifacts are as expected.
+    assert.exists(signedTransaction)
+    assert.deepEqual(signedTransaction, expectedSignedTransaction)
+  })
+
+  it('Sign TrustSet transaction with no limitAmount', function (): void {
+    // GIVEN a Payment transaction without a limitAmount field and a wallet.
```
amiecorso, line 441:
```suggestion
    // GIVEN a TrustSet transaction without a limitAmount field and a wallet.
```

amiecorso (approved):
Looks great, just a few comments.  Thanks for doing this!! now we can bump the codec and sleep well at night.

tedkalaw (approved):

PR #569, opened 2020-10-20
Serializes flags on transactions

tedkalaw:
This PR serializes the optional `Flags` field on transactions, and adds a test.

### Context of Change
In order to authorize a Trust Line, we need to set a `Flag` on a `TrustSet` transaction. Unfortunately, we are not currently serializing this field.

### Type of Change

- [ ] Documentation Updates
- [ ] Release

## Before / After
Flags are serialized instead of silently being ignored

## Test Plan
CI, the test I added

### DIFF
--- a/test/XRP/serializer.test.ts
+++ b/test/XRP/serializer.test.ts
@@ -2712,7 +2712,7 @@ describe('serializer', function (): void {
     assert.deepEqual(serialized, expectedJSON)
   })
 
-  it.only('serializes a TrustSet transaction with flags set', function (): void {
+  it('serializes a TrustSet transaction with flags set', function (): void {
     // GIVEN a transaction which represents the creation of a trust line linking two accounts.
     const currency = 'USD'
     const currencyIssuer = 'XVPcpSm47b1CZkf5AkKM9a84dQHe3m4sBhsrA4XtnBECTAc'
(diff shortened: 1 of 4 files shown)


test/XRP/serializer.test.ts
```diff
@@ -2710,6 +2712,47 @@ describe('serializer', function (): void {
     assert.deepEqual(serialized, expectedJSON)
   })
 
+  it('serializes a TrustSet transaction with flags set', function (): void {
+    // GIVEN a transaction which represents the creation of a trust line linking two accounts, with flags.
+    const currency = 'USD'
+    const currencyIssuer = 'XVPcpSm47b1CZkf5AkKM9a84dQHe3m4sBhsrA4XtnBECTAc'
+    const transaction = xrpTestUtils.makeTrustSetTransaction(
```
amiecorso, line 2719: should we just update this method to optionally take flags as an argument?  Maybe not, I don't know, just an idea

amiecorso (approved):
just the one idea for keeping our test utils up to date but otherwise: 
eating-tide-pods

test/XRP/serializer.test.ts
tedkalaw: i did it, seems reasonable

mvadari (approved):

PR #604, opened 2020-11-20
Fixes A Bug In PathElement Serialization

amiecorso:
This PR fixes a bug in the serializer code that incorrectly assigned the field `currencyCode` to a path element instead of the `currency` field expected by the rippled server.  We have signer tests for these serialized objects that do cover paths, but this bug wasn't obviously detectable until attempting to send an actual transaction to the network.

### Type of Change

- [x] Bug fix (non-breaking change which fixes an issue)

## Before / After
Path elements correctly serialized.

## Test Plan
Change tests to check for correct `currency` field.  npm link locally to test successful submission of transaction with paths.

### DIFF
--- a/src/XRP/serializer.ts
+++ b/src/XRP/serializer.ts
@@ -324,7 +324,7 @@ interface MemoDetailsJSON {
 interface PathElementJSON {
   account?: string
   issuer?: string
-  currencyCode?: CurrencyJSON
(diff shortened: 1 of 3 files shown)


amiecorso:
I've never seen codecov this color before

mvadari (approved):

tedkalaw (approved):

